PR #33, opened 2024-08-02
Macros for maps and sets

### LINKED ISSUE #28: Macros for other collections
Hi, `bon` looks really nice. Thank you for the project.

In the docs, you mention wanting to add `vec`-like macros for other collections. Would you be interested in a PR for that?

korrat:
Fixes #28.

Adds `map` and `set` macros, which create any map or set type from an iterator of `(K, V)` pairs or `V`, respectively.

Both macros perform `Into::into` conversion on their arguments and perform a rudimentary uniqueness check (syntax only).

I created multiple commits while figuring things out. If you prefer, I can squash them down to a single commit.


Cargo.toml
```diff
@@ -139,6 +139,7 @@ wildcard_dependencies             = "warn"
 zero_sized_map_values             = "warn"
 
 [workspace.lints.rustdoc]
+broken_intra_doc_links   = "warn"
```
Veetaha, line 142: Here we enable lints that are `allow` by default because there tons of cool and useful lints in rustc, clippy and rustdoc that aren't enabled by default due to various reasons.

This lint, however, is already [`warn` by default](https://doc.rust-lang.org/rustdoc/lints.html#broken_intra_doc_links), so I don't think we need to list it here explicitly.

bon-macros/src/lib.rs
```diff
@@ -100,3 +102,93 @@ pub fn bon(params: TokenStream, item: TokenStream) -> TokenStream {
...
         .into()
 }
+
+/// Similar to [`maplit::hashmap!`]/[`maplit::btreemap!`] but converts each key and value with
+/// [`Into::into()`].
+///
+/// There are no separate variants for [`HashMap`] and [`BTreeMap`] since the macro works with any
+/// type that implements [`FromIterator<(K, V)>`].
+///
+/// The macro also performs rudimentary uniqueness checking on keys: Syntactically equal keys are
```
Veetaha, line 112:
```suggestion
/// The macro also performs rudimentary uniqueness checking on keys: syntactically equal keys are
```

bon-macros/src/lib.rs
```diff
@@ -100,3 +102,93 @@ pub fn bon(params: TokenStream, item: TokenStream) -> TokenStream {
...
+///     let ffmpeg_args: HashMap<String, String> = bon::map!{
+///         "-i": format!("input.{input_extension}"),
+///         "-y": format!("output.{output_extension}"),
+///     };
+///
+///     let mut command = std::process::Command::new("ffmpeg");
+///     for (flag, value) in ffmpeg_args {
+///         command.arg(flag).arg(value);
+///     }
+///     command.output()?;
```
Veetaha, line 132: There are a couple of problems with this example.

First, it should showcase that both the keys of the map can have diverse types and values too. For example:

```rust
let map: HashMap<String, String> = bon::map! { 
    "&str literal key": format!("Owned string value: {}", 2 + 2),
    format!("Owned String key"): "&str literal value",
};
```

I can't give an immediate good real-world example to showcase this, but the current example with `ffmpeg` CLI doesn't show this feature.

The second problem is a bit of a logical bug. The flag `-y` in `ffmpeg` is  a boolean fla. From the `ffmpeg` docs: (https://www.ffmpeg.org/ffmpeg.html)):


So it doesn't take any value, and it isn't correct to pair it with the output file path in the map key-value pair.

If you can't think of any other good real-world use case for a map with some keys of string literal types, and some of dynamic owned strings, then let's just use a dummy example that uses `foo`s and `bar`s as values. We can update the docs with a better example any time anyway.

bon-macros/src/lib.rs
```diff
@@ -100,3 +102,93 @@ pub fn bon(params: TokenStream, item: TokenStream) -> TokenStream {
...
+        .into()
+}
+
+/// Similar to [`maplit::hashset!`]/[`maplit::btreeset!`] but converts each key and value with
+/// [`Into::into()`].
+///
+/// There are no separate variants for [`HashSet`] and [`BTreeSet`] since the macro works with any
+/// type that implements [`FromIterator<T>`].
+///
+/// The macro also performs rudimentary uniqueness checking on keys: Syntactically equal keys are
```
Veetaha, line 157:
```suggestion
/// The macro also performs rudimentary uniqueness checking on values: syntactically equal keys are
```

bon-macros/src/lib.rs
```diff
@@ -100,3 +102,93 @@ pub fn bon(params: TokenStream, item: TokenStream) -> TokenStream {
...
+/// rejected with a compile error.
+///
+/// A good example of the use case for this macro is when you want to create a
+/// `Hashset<String, String>` where part of the keys or values are hardcoded string literals of type `&str`
```
Veetaha, line 161:
```suggestion
/// `HashSet<String>` where part of the values are hardcoded string literals of type `&str`
```

bon-macros/src/lib.rs
```diff
@@ -100,3 +102,93 @@ pub fn bon(params: TokenStream, item: TokenStream) -> TokenStream {
...
+/// ```rust
+/// # use bon_macros as bon;
+/// # use std::collections::HashSet;
+/// fn convert_media(input_extension: &str, output_extension: &str) -> std::io::Result<()> {
+///     let ffmpeg_args: HashSet<String> = bon::set!{
+///         "-i",
+///         format!("input.{input_extension}"),
+///         "-y",
+///         format!("output.{output_extension}"),
+///     };
```
Veetaha, line 173: This example is also not correct. `HashSet` produces elements in arbitrary order. It doesn't make sense to put CLI args in a set anyway.

bon-macros/src/set.rs
```diff
@@ -0,0 +1,21 @@
...
+
+    let items = entries.into_iter().map(|value| {
+        let value = quote!(::core::convert::Into::into(#value));
+        value
+    });
+
+    Ok(quote! {
+        ::core::iter::FromIterator::from_iter([
+            #(#items),*
+        ])
```
Veetaha, line 19: This bit can be simplified:

```suggestion
    Ok(quote! {
        ::core::iter::FromIterator::from_iter([
            #(::core::convert::Into::into(#items)),*
        ])
```

bon-macros/src/util/mod.rs
```diff
@@ -45,6 +49,39 @@ where
...
+    keys.into_iter().for_each(|key| {
+        if !exprs.insert(key.clone()) {
+            errors.push(err!(key, "duplicate map key"));
+        }
+    });
+
+    errors.finish()?;
+
+    Ok(())
+}
```
Veetaha, line 83: Note that this check should not include expressions that include function or method calls, because functions are not guaranteed to be pure.

For example, in this check will reject the following valid code that would produce unique set items:

```rust
let i = 0;

let mut incr = || {
    i += 1;
    i
};

let set = bon::set![
    incr(),
    incr(),
    incr(),
];
```
In this case the set would've contained the values `1`, `2` and `3`, but the macro would reject it.

Instead we need to check only those expressions that consist of primitive literals `syn::ExprLit` and operators `+`, `-`, `==`, `!=`, etc. (not function calls).

bon/src/lib.rs
```diff
@@ -95,4 +96,52 @@ mod tests {
...
+
+    #[test]
+    fn set_smoke() {
+        use std::collections::BTreeSet;
+        use std::collections::HashSet;
+
+        let hash_strings: HashSet<String> = crate::set! {
+            "Hello", "World",
+            "Goodbye", "Mars",
+        };
```
Veetaha, line 130: Let's use square brackets and propagate square brackets syntax for set in the docs as well. I believe `set![]` is closer to a `vec![]` in this regard.

```suggestion
        let hash_strings: HashSet<String> = crate::set![ 
            "Hello", "World",
            "Goodbye", "Mars",
        ];
```

Cargo.toml
korrat: Sorry, missed that part of the docs. Fixed.

bon-macros/src/lib.rs
korrat: I did a mostly synthetic example that still shows the macro in action.

I could also split the example, showing the working code first and a compile fail second.

bon-macros/src/lib.rs
korrat: Similar synthetic example.

bon-macros/src/util/mod.rs
korrat: Didn't think of that one. We're now only checking for “pure” expressions (naming subject to bikeshedding) for uniqueness.

korrat:
I've made the requested changes as a series of fixup commits. Let me know when you consider this PR ready to merge, then I can rebase & squash them down.

bon-macros/src/lib.rs
Veetaha: Let's probably avoid using `ffmpeg` in the examples. Let's make the either fully real-world, or fully synthetic. It could look like a unit test that just creates a map and asserts that it contains the expected keys. And yeah, a separate example for the validation of uniqueness would be cool

bon-macros/src/util/mod.rs
```diff
@@ -45,6 +49,52 @@ where
...
+
+fn is_pure(expr: &Expr) -> bool {
+    match expr {
+        Expr::Binary(binary) => is_pure(&binary.left) && is_pure(&binary.right),
+        Expr::Lit(_) => true,
+        Expr::Paren(paren) => is_pure(&paren.expr),
```
Veetaha, line 92: There is also [`Expr::Group`](https://docs.rs/syn/latest/syn/token/struct.Group.html), which is similar to `Paren`, but it's a bit different beast. It is an expression delimited by "invisible" parens. This token can only be created from macro invocations, but we may include its handling here as well similar to `Expr::Parent` logic

bon/tests/integration/ui/compile_fail/map_macro.stderr
```diff
@@ -0,0 +1,13 @@
...
+14 |         "Hello": "Mars",
+   |         ^^^^^^^
+
+warning: unused variable: `strings`
+ --> tests/integration/ui/compile_fail/map_macro.rs:5:9
+  |
+5 |     let strings: HashMap<u64, String> = map! {
+  |         ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_strings`
+  |
+  = note: `#[warn(unused_variables)]` on by default
```
Veetaha, line 13: Let's add `_` to the variable name to avoid having this warningin the stderr snapshot

bon/tests/integration/ui/compile_fail/set_macro.stderr
```diff
@@ -0,0 +1,13 @@
...
+14 |         "Hello", "Mars",
+   |         ^^^^^^^
+
+warning: unused variable: `strings`
+ --> tests/integration/ui/compile_fail/set_macro.rs:5:9
+  |
+5 |     let strings: HashSet<u64> = set! {
+  |         ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_strings`
+  |
+  = note: `#[warn(unused_variables)]` on by default
```
Veetaha, line 13: Ditto

Veetaha (commented):
Let's also update this paragraph on the website:
https://github.com/elastio/bon/blob/cfd50943d1b02dc0032a978739f9ed9ab09341f1/website/docs/guide/alternatives.md?plain=1#L64

I think these are basically all the macros that we need in `bon` for collections at least for now.

Also, you don't need to squash your commits. Github can do it automatically when I click the merge button. The PR title will become the squashed commit message.

bon/tests/integration/ui/compile_fail/map_macro.stderr
korrat: Fixed by adding a `compile_fail` doctest instead.

Veetaha:
Thank you for the PR! I'll prepare and cut a minor release in the next couple of days

Veetaha (approved):

PR #36, opened 2024-08-05
add no_std support

danielschemmel:
I wanted to try out `bon` in one of my `no_std` crates, but it turns out that `bon` does not have `no_std` support.

As a generator for builders, I could not see a reason why it would not have that, so here is a stab at adding `no_std` support. The example works for me, although I am sure I missed a spot or two ;).

Implementation was straightforward, with only `bon::vec!` requiring any thought at all, as the `alloc` crate is not available by default in either `std` or `no_std` configurations. I went the standard route of adding two features, `std` and `alloc`, with `std` enabled by default.
- If `std` is enabled, `bon::vec!` works as before and uses `std::vec`
- If `alloc` is enabled and `std` is not, `bon::vec!` uses `alloc::vec`
- If neither are enabled, `bon::vec!` is not available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,8 @@ jobs:
 
       - run: cargo test --locked --all-features --all-targets
       - run: cargo test --locked --all-features --doc
+      - run: cd bon && cargo test --locked --no-default-features --features=
+      - run: cd bon && cargo test --locked --no-default-features --features=alloc
 
   cargo-clippy:
     runs-on: ${{ matrix.os }}-latest
--- a/bon/src/lib.rs
+++ b/bon/src/lib.rs
@@ -33,6 +33,9 @@ pub mod private;
 ///
 /// This macro doesn't support `vec![expr; N]` syntax, since it's simpler to
 /// just write `vec![expr.into(); N]` using [`std::vec!`] instead.
+///
+/// This macro is only available if the `std` or the `alloc` feature is enabled. The
+/// `std` feature is enabled by default.
 #[macro_export]
 #[cfg(feature = "alloc")]
 macro_rules! vec {
--- a/bon/src/lib.rs
+++ b/bon/src/lib.rs
@@ -20,9 +20,9 @@ pub mod private;
 /// fn convert_media(input_extension: &str, output_extension: &str) -> std::io::Result<()> {
 ///     let ffmpeg_args: Vec<String> = bon::vec![
 ///         "-i",
-///         alloc::format!("input.{input_extension}"),
+///         format!("input.{input_extension}"),
 ///         "-y",
-///         alloc::format!("output.{output_extension}"),
+///         format!("output.{output_extension}"),
 ///     ];
 ///
 ///     std::process::Command::new("ffmpeg").args(ffmpeg_args).output()?;
@@ -79,7 +79,9 @@ macro_rules! arr {
 }
 
 #[cfg(test)]
+#[cfg(feature = "alloc")]
 mod tests {
+    use crate::private::alloc::{string::String, vec::Vec};
 
     #[test]
     fn arr_smoke() {
@@ -99,6 +101,7 @@ mod tests {
         assert!(actual.is_empty());
     }
 
+    #[cfg(feature = "std")]
     #[test]
     fn map_smoke() {
         use std::collections::BTreeMap;
@@ -121,6 +124,7 @@ mod tests {
         assert_eq!(tree_strings["Goodbye"], "Mars");
     }
 
+    #[cfg(feature = "std")]
     #[test]
     fn set_smoke() {
         use std::collections::BTreeSet;
--- a/bon/src/lib.rs
+++ b/bon/src/lib.rs
@@ -1,4 +1,5 @@
 #![doc = include_str!("../README.md")]
+#![cfg_attr(not(feature = "std"), no_std)]
 
 pub use bon_macros::*;
 
@@ -19,9 +20,9 @@ pub mod private;
 /// fn convert_media(input_extension: &str, output_extension: &str) -> std::io::Result<()> {
 ///     let ffmpeg_args: Vec<String> = bon::vec![
 ///         "-i",
-///         format!("input.{input_extension}"),
+///         alloc::format!("input.{input_extension}"),
 ///         "-y",
-///         format!("output.{output_extension}"),
+///         alloc::format!("output.{output_extension}"),
 ///     ];
 ///
 ///     std::process::Command::new("ffmpeg").args(ffmpeg_args).output()?;
@@ -33,9 +34,10 @@ pub mod private;
 /// This macro doesn't support `vec![expr; N]` syntax, since it's simpler to
 /// just write `vec![expr.into(); N]` using [`std::vec!`] instead.
 #[macro_export]
+#[cfg(feature = "alloc")]
 macro_rules! vec {
-    () => (::std::vec::Vec::new());
-    ($($item:expr),+ $(,)?) => (::std::vec![$(::core::convert::Into::into($item)),+ ]);
+    () => ($crate::private::alloc::vec::Vec::new());
+    ($($item:expr),+ $(,)?) => ($crate::private::alloc::vec![$(::core::convert::Into::into($item)),+ ]);
 }
 
 /// Creates a fixed-size array literal where each element is converted with [`Into::into()`]
--- a/bon/tests/integration/builder_on_fn.rs
+++ b/bon/tests/integration/builder_on_fn.rs
@@ -1,9 +1,13 @@
 mod expose_positional_fn;
 
 use bon::{bon, builder};
-use std::collections::BTreeSet;
-use std::num::NonZeroU32;
+#[cfg(feature = "alloc")]
+use {
+    alloc::borrow::ToOwned, alloc::collections::BTreeSet, alloc::format, alloc::string::String,
+    alloc::vec, alloc::vec::Vec, core::num::NonZeroU32,
+};
 
+#[cfg(feature = "alloc")]
 #[test]
 fn smoke() {
     /// Function-level docs
@@ -51,6 +55,7 @@ fn smoke() {
     assert_eq!(actual, "arg3");
 }
 
+#[cfg(feature = "alloc")]
 #[test]
 fn default_attr() {
     #[builder]
@@ -68,6 +73,7 @@ fn default_attr() {
     assert_eq!(actual, (0, 42, "default".to_owned(), vec![42]));
 }
 
+#[cfg(feature = "alloc")]
 #[test]
 fn into_attr() {
     #[builder]
@@ -102,6 +108,7 @@ fn into_attr() {
     assert_eq!(actual, "vinyl-scratch:32:Some({32, 43}):disabled");
 }
 
+#[cfg(feature = "alloc")]
 #[test]
 fn into_string() {
     #[builder]
@@ -137,6 +144,7 @@ fn lifetime_elision() {
     assert_eq!(actual, ("blackjack", "blackjack", ["blackjack"]));
 }
 
+#[cfg(feature = "std")]
 #[tokio::test]
 async fn async_func() {
     #[builder]
@@ -162,13 +170,14 @@ fn unsafe_func() {
     unsafe { builder.call() };
 }
 
+#[cfg(feature = "alloc")]
 #[test]
 fn impl_traits() {
     #[builder]
     fn sut(
         /// Some documentation
         iterable: impl IntoIterator<Item = impl Into<u32>>,
-        showable: impl std::fmt::Display + std::fmt::Debug,
+        showable: impl core::fmt::Display + core::fmt::Debug,
     ) -> (String, Vec<u32>) {
         let str = format!("{showable} + {showable:#?}");
         let vec = iterable.into_iter().map(Into::into).collect();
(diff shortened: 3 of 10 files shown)


Veetaha:
Hi! Thank you for the PR. `no_std` support is definitely something we can have in `bon`, that's why I've been preferring `::core` references to `::std`, although not that strictly (since the PR diff). I, however, didn't have any experience with `no_std` in Rust. I'll need some time to read the docs and understand the difference between `std`, `core`, and `alloc` to be able to maintain `no_std` support from now one.

Also, to maintain the `no_std` support I think we need some tests on CI to make sure that no `::std` references are used in the generated code. For example, if I add some new features to the proc macro I'd like to know that `no_std` support wasn't broken by the new feature.

Maybe we could add some new crate to the workspace e.g. `no-std-tests` where we add `bon` as a dependency with `no-deafult-features = true`, and make sure some basic functionality of it works. 

Another thing.. We should probably have something mentioned in the docs about `no_std` support. Maybe it could be shortly mentioned in the main overview in a `::: tip` element or smth. See the `docs/guid/overview.md` in the `website` directory.

bon/src/lib.rs
```diff
@@ -33,11 +34,45 @@ pub mod private;
...
+/// }
+/// ```
+///
+/// This macro doesn't support `vec![expr; N]` syntax, since it's simpler to
+/// just write `vec![expr.into(); N]` using [`std::vec!`] instead.
+#[macro_export]
+#[cfg(all(feature = "alloc", not(any(test, feature = "std"))))]
+macro_rules! vec {
+    () => (::alloc::vec::Vec::new());
+    ($($item:expr),+ $(,)?) => (::std::vec![$(::alloc::convert::Into::into($item)),+ ]);
```
Veetaha, line 73: I need to dig a bit more into this, but my immediate reaction is "why does this macro have to be duplicated"? Is there a way to avoid this? Shouldn't `std` automatically imply `alloc` is available?

danielschemmel: I absolutely agree that it should!

[But it is not](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=45b502af665cd5afd3dcbac59699b783)

Veetaha:
Another thing. By looking at `serde` crate, I see that it adds `no_std` to `categories` and `keywords`: https://github.com/serde-rs/serde/blob/e08c5de5dd62571a5687f77d99609f9d9e60784e/serde/Cargo.toml#L6-L11.

I think we can do that as well to improve discoverability.

bon/src/lib.rs
Veetaha: So the problem is that `vec/Vec` lives in `alloc` crate, which isn't accessible by default? That's really inconvenient.. I'm sure other Rust crates have some good ways of fighting with this..

I took a look at `serde`, and here is [what it does](https://github.com/serde-rs/serde/blob/e08c5de5dd62571a5687f77d99609f9d9e60784e/serde/src/lib.rs#L211-L214).

I've also been thinking of using a conditionally-compiled alias. We can just add aliases like this to `private` module (which is `pub` but `doc(hidden)`, and references them in the macro as `$crate::private::Vec/vec`

Veetaha: Btw. serde also adds `extern crate alloc` under the `alloc` feature:
https://github.com/serde-rs/serde/blob/e08c5de5dd62571a5687f77d99609f9d9e60784e/serde/src/lib.rs#L161-L162

danielschemmel:
Great to hear that you are open to this idea! In a nutshell: `core` is everything that does not rely on the OS in any way. `alloc` adds those parts that only rely on dynamic memory allocation. `std` then adds everything else, such as files or networking. Many crates should be able to rely on just `core` and `alloc`!

I have added the categories, but not the keywords, as [crates.io limits keywords to a total of 5](https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field).

As a discussion position, I have also copied the integration tests to `nostd-tests`, where they should be executed in the correct manner. I edited the tests to remove anything that I could not immediately figure out how to do in an equivalent (enough) manner with only core types. If this works for you, I can do the same thing for an `alloc` version.

Note that features of all dependencies in a workspace are aggregated (that is arguably the point of a workspace, after all), so it makes little sense to put this crate into the workspace, as it would pick up the `std` features of the other crates.

danielschemmel:
I can also write a section for the docs/webpage if/once we agree on the general layout.

Veetaha:
As for features - that's true, but we can run tests separately with different combinations of features, which is also important to test (e.g. see [serde's ci](https://github.com/serde-rs/serde/blob/e08c5de5dd62571a5687f77d99609f9d9e60784e/.github/workflows/ci.yml#L61-L66)).

I'll take a look at the changes a bit later, and maybe polish some things from my side once this lands. I hope to get this landed and released together with other unrelated PRs this week

Veetaha:
> as it would pick up the std features of the other crates.

It won't if cargo command such as `cargo test` selects a subset of packages `-p nostd-tests` for example. In this case only features enabled by this crate will be used

Veetaha:
> I have also copied the integration tests to nostd-tests

Maybe we could have tests in the same crate but under the cfg feature gates. For example nostd compatible test will have no cfgs on them. Then alloc or std tests will be under modules that have cfg feature flags for std and alloc.

Then CI should run tests with different combinations of selected features

danielschemmel:
I can use the existing tests, it seems as if `cargo test --no-default-features` gets picked up as intended, _when run in `/bon`, not the workspace.

danielschemmel:
I'll need some time to go over all the tests and match them to the three levels.

Veetaha:
Yep let's do that 👍 . Serde is a good example to follow in this regard

danielschemmel:
I did a very quick pass where I just disabled everything that is not nostd/alloc ready, and added it to the CI.

Does this fit your idea?

Veetaha:
Yeah, I'd also consider moving tests under child modules std.rs, alloc.rs and no_std.rs where cfg attributes are then written on top of the mod declarations. This will avoid adding cfg attributes on each test function

bon/src/lib.rs
danielschemmel: Yes, so do we, but just for tests. The crate is not exported (which it is also not in serde).

One other way of doing things that would avoid the duplication is to `pub extern crate alloc;` (Without the `pub extern`, you cannot `pub use` it - and once it is `pub extern`, there is no need to `pub use` it.). Would you like me to push an example?

Veetaha: pub extern crate works fine as well. Everything inside of the private module is considered private, eo feelf free to pub {anything} ther3

Veetaha: I see you added it at the crate root. Is this the limitation of `extern crate` that it needs to be at the crate root? Anyway this way is also fine as long as that symbol is `doc(hidden)`

danielschemmel: Done (and moved to private)

danielschemmel:
Alright, I did another pass over the tests, and it turns out only a few of them really need `alloc` , and only the `map_smoke` and `set_smoke` tests use `std` at all. Some of the tests that are still restricted might lose that restriction if someone who knows what the real point of it where to look at it ;).

I don't think the tests would benefit from separating out the `std` tests, at this point.

danielschemmel:
And I also added a small doc blurb at the location you suggested.

.github/workflows/ci.yml
```diff
@@ -41,6 +41,8 @@ jobs:
 
       - run: cargo test --locked --all-features --all-targets
       - run: cargo test --locked --all-features --doc
+      - run: cd bon && cargo test --locked --no-default-features --features=
+      - run: cd bon && cargo test --locked --no-default-features --features=alloc
```
Veetaha, line 45: Let's include doc tests in these test runs
```suggestion
      - run: cd bon && cargo test --locked --all-targets --doc --no-default-features --features=
      - run: cd bon && cargo test --locked --all-targets --doc --no-default-features --features=alloc
```

bon/tests/integration/builder_on_fn.rs
```diff
@@ -26,29 +29,27 @@ fn smoke() {
 
         /// Docs for arg2
         arg2: &'_ str,
-        arg3: String,
+        arg3: u64,
```
Veetaha, line 32: Let's keep the smoke tests under `alloc` feature. They validate into conversions for `String`.

bon/tests/integration/builder_on_fn.rs
```diff
@@ -57,17 +58,18 @@ fn default_attr() {
     fn sut(
         #[builder(default)] arg1: u32,
         #[builder(default = 42)] arg2: u32,
-        #[builder(default = "default")] arg3: String,
-        #[builder(default = vec![42])] arg4: Vec<u32>,
-    ) -> (u32, u32, String, Vec<u32>) {
+        #[builder(default = "default")] arg3: &str,
```
Veetaha, line 61: This should also stay under `alloc`. It validates that `default` close does `into` conversion on its expression. Alterantively, these last two arguments could be separated into a different test under `alloc`, to keep the test for the first two arguments no_std-compatible

Veetaha (commented):
I think, to save some time, I'll go over the tests from my side and do the changes to them according to the context. For now, we can merge it like this with the comments I mentioned fixed

Veetaha:
I've merged my PR #29 which introduced some small merge conflicts here.

bon/src/lib.rs
```diff
@@ -78,16 +83,19 @@ macro_rules! arr {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "alloc")]
+    use crate::private::alloc::{string::String, vec::Vec};
 
     #[test]
     fn arr_smoke() {
-        let actual: [String; 3] = crate::arr!["foo", "bar", "baz"];
+        let actual: [&str; 3] = crate::arr!["foo", "bar", "baz"];
```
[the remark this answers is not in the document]
danielschemmel: Done - since all tests here use at least alloc, I moved the cfg for that to the module level as well.

bon/src/lib.rs
```diff
@@ -16,26 +17,30 @@ pub mod private;
...
+/// # extern crate alloc;
+/// # use alloc::{string::String, vec::Vec};
+/// fn convert_media(input_extension: &str, output_extension: &str) -> Vec<String> {
 ///     let ffmpeg_args: Vec<String> = bon::vec![
 ///         "-i",
-///         format!("input.{input_extension}"),
+///         alloc::format!("input.{input_extension}"),
 ///         "-y",
-///         format!("output.{output_extension}"),
+///         alloc::format!("output.{output_extension}"),
```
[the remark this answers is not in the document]
danielschemmel: I am fairly certain I caused it to throw up issues at some time, but it is not an issue anymore. I reverted the changes to that doc-test.

.github/workflows/ci.yml
danielschemmel: I think that doc tests are run by default:
```
❭ cd bon && cargo test --locked --no-default-features --features=
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running unittests src/lib.rs (/home/danielschemmel/projects/bon/target/debug/deps/bon-0551bf49c71843ca)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/integration/main.rs (/home/danielschemmel/projects/bon/target/debug/deps/integration-d2abf59b71eb5bec)

running 22 tests
test builder_on_fn::const_function ... ok
test builder_on_fn::expose_positional_fn::method_new_doesnt_require_a_value_for_name ... ok
test builder_on_fn::constructor ... ok
test builder_on_fn::expose_positional_fn::simple ... ok
test builder_on_fn::expose_positional_fn::with_nested_params ... ok
test builder_on_fn::generics_with_lifetimes ... ok
test builder_on_fn::impl_block_ty_contains_a_reference ... ok
test builder_on_fn::impl_block_with_self_in_const_generics ... ok
test builder_on_fn::leading_underscore_is_stripped ... ok
test builder_on_fn::lifetime_elision ... ok
test builder_on_fn::mut_fn_params ... ok
test builder_on_fn::raw_identifiers ... ok
test builder_on_fn::receiver ... ok
test builder_on_fn::receiver_is_non_default ... ok
test builder_on_fn::receiver_with_lifetimes ... ok
test builder_on_fn::self_in_a_bunch_of_places ... ok
test builder_on_fn::self_only_generic_param ... ok
test builder_on_fn::types_not_implementing_default ... ok
test builder_on_fn::unsafe_func ... ok
test builder_on_struct::raw_identifiers ... ok
test builder_on_struct::smoke ... ok
     Locking 1 package to latest compatible version
      Adding bon-tests v0.0.0 (/home/danielschemmel/projects/bon/target/tests/trybuild/bon)
    Checking bon-tests v0.0.0 (/home/danielschemmel/projects/bon/target/tests/trybuild/bon)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.07s


test tests/integration/ui/compile_fail/misc.rs ... ok


test ui::ui ... ok

test result: ok. 22 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.18s

   Doc-tests bon

running 2 tests
test bon/src/lib.rs - arr (line 58) ... ok
test bon/src/lib.rs - (line 23) ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.09s
```

The last two tests are the doc tests (this run is nostd without alloc, so there are only two doc tests).

Either way, your suggested command does not have the intended result:
```
❭ cd bon && cargo test --locked --all-targets --doc --no-default-features --features=
error: Can't mix --doc with other target selecting options
```

bon/tests/integration/builder_on_fn.rs
danielschemmel: Done

bon/tests/integration/builder_on_fn.rs
danielschemmel: Done

danielschemmel:
I rebased on the current master branch.

.github/workflows/ci.yml
Veetaha: Aha, makes sense

Veetaha:
Thank you for the pull request! I'll try to get it released tomorrow

Veetaha (approved):

PR #54, opened 2024-08-19
`#[builder(on(type_pattern, attrs))]`

### LINKED ISSUE #15: Is changing `String` to `Into<String>` correct?
When using `#[builder]` on a function, `String` arguments are turned into `Into<String>`. This feels wrong — `String` may have been an intentional choice by the author to, for instance, force allocations to be easily visible. Changing it to `Into<String>` defeats the purpose of that.

I think it's perfectly fine, and even correct, for callers to have to call `.into()` on their strings, just as if they had called the function using normal Rust syntax.

Veetaha:
Closes #15. See details in https://github.com/elastio/bon/issues/15#issuecomment-2297202745


anelson (approved):

Veetaha:
💥

PR #76, opened 2024-08-29
docs: improve builder() method docs

EdJoPaTo:
When I use an API I first care about the what, then how to achieve it. So reversing the statement feels better to me.

### DIFF
--- a/bon-macros/src/builder/builder_gen/input_struct.rs
+++ b/bon-macros/src/builder/builder_gen/input_struct.rs
@@ -132,7 +132,7 @@ impl StructInputCtx {
         };
 
         let start_func_docs = format!(
-            "Use builder syntax to create an instance of [`{}`]",
+            "Create an instance of [`{}`] using builder syntax",
             self.norm_struct.ident
         );
 


bon-macros/src/builder/builder_gen/input_struct.rs
```diff
@@ -132,7 +132,7 @@ impl StructInputCtx {
         };
 
         let start_func_docs = format!(
-            "Use builder syntax to create an instance of [`{}`]",
+            "Create an instance of [`{}`] using builder syntax",
```
EdJoPaTo, line 135: Alternative:
```suggestion
            "Create an instance of [`{}`] with builder syntax",
```

Veetaha: Hi! Thank you for opening a PR! I think this makes sense in this case.

Generally, I try to put the dynamic part of the message as the end. This way the static text around the interpolated value always stays at the beginning and is completed by the time the person is reading the dynamic part.

A similar problem was in rust's panic messages [before 1.73.0, where they changed it](https://blog.rust-lang.org/2023/10/05/Rust-1.73.0.html#cleaner-panic-messages). In the older version of Rust you had to read the entire panic message first (which could take multiple lines actually), and only after that you'd read the line number (which is always a static limited size of text) at the end:

```
thread 'main' panicked at 'oh no! "ferris.txt" not found!', src/main.rs:3:5
```

Imagine reading "thread 'main' panicked at '{ huge debug formatted output by the end of which you forgot the start of the sentence }, src/main.rs:3:5'. It becomes harder to find that line reference and jump to it.

In this case there shouldn't be such a problem, because we interpolate a single identifier in the docs message, so I agree with this change, and I'll merge the PR once CI finishes.

I don't think it's critical to release this change immediately unless you think otherwise. I'd release it with some other fixes that I'm planning to do (e.g. cfg/cfg_attr conditional compilation support).

EdJoPaTo: I think the case of the rust panic is different, as the panic message is usually quite long.

Veetaha (approved):

PR #79, opened 2024-08-29
docs: doc-markdown and some typos

EdJoPaTo:
Noticed a typo, checked with cargo-spellcheck and added [`clippy::doc_markdown`](https://rust-lang.github.io/rust-clippy/master/index.html#/doc_markdown) as a lint.


bon-macros/src/util/ty/mod.rs
```diff
@@ -3,7 +3,7 @@ mod match_types;
 use crate::util::prelude::*;
 
 pub(crate) trait TypeExt {
-    /// Try downcasting the type to [`syn::Type::Path`]
+    /// Try down-casting the type to [`syn::Type::Path`]
```
Veetaha, line 6: I believe rust's style is to write `downcast`. See how it's spelled in [std docs](https://doc.rust-lang.org/stable/std/boxed/struct.Box.html#method.downcast-2)

Veetaha (commented):
I wonder how `cargo-spellcheck` compares to [`typos`](https://github.com/crate-ci/typos), which is currently used on CI. Does it make sense to add `cargo-spellcheck` to CI as well? How big is the rate of false-positives there?

EdJoPaTo:
cargo-spellcheck creates a bunch of false positives pretty quickly but seems to checks different places than `typos`. Stuff like impl or struct is written sometimes in the docs and cargo-spellcheck isn't happy about it. Personally I haven't added it to my CI stuff and only use it on demand as its just simpler to go through them when relevant. It's also not as good as something like LanguageTool (and I don't know a tool which works with Rust code there)

Veetaha:
Okay, let's keep it out of CI for now. I also use Grammarly although painfully manually to spellcheck the markdown files in the `website` directory. They discontinued their support for the SDK that their unofficial VSCode extension was using..

Veetaha (approved):

EdJoPaTo:
I use [LTeX](https://open-vsx.org/vscode/item?itemName=valentjn.vscode-ltex) on VSCode with LanguageTool installed on the same device for Markdown checking which works well enough for me currently.

Veetaha:
Thanks! I'll check it out

PR #80, opened 2024-08-29
fix(clippy): don't warn on clippy::impl_trait_in_params

### LINKED ISSUE #78: into triggers `clippy::impl_trait_in_params`
See https://rust-lang.github.io/rust-clippy/master/index.html#/impl_trait_in_params.

Reproduce:

```rust
#![warn(clippy::impl_trait_in_params)]

fn main() {
    let example = Example::builder().content(42).build();
    dbg!(example.content);
}

#[bon::builder]
pub struct Example {
    #[builder(into)]
    content: i32,
}
```

Maybe it's a good idea to follow this lints suggestion and move to `fn content<T: Into<i32>(self, value: T)`. Otherwise, maybe ignore this lint by default (like other non clippy lints are already ignored)

EdJoPaTo:
fixes #78

Thought about adding [`unknown_lints`](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#unknown-lints) too, but [`clippy::impl_trait_in_params` was added in 1.69](https://rust-lang.github.io/rust-clippy/master/index.html#/impl_trait_in_params) and the current MSRV seems to be 1.70 due to usage of [`is_some_and`](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and) (and similar methods) so `unknown_lints` can't happen for this lint.

### DIFF
--- a/bon-macros/src/builder/builder_gen/setter_methods.rs
+++ b/bon-macros/src/builder/builder_gen/setter_methods.rs
@@ -271,6 +271,7 @@ impl<'a> MemberSettersCtx<'a> {
 
         quote! {
             #( #docs )*
+            #[allow(clippy::impl_trait_in_params)]
             #[inline(always)]
             #vis fn #method_name(self, #fn_params) -> #return_type {
                 #builder_ident {


Veetaha:
I haven't settled on a specific MSRV yet though... I've been postponing this work all the time. I've just created an issue about it https://github.com/elastio/bon/issues/81.

Veetaha (approved):

PR #82, opened 2024-08-30
feat: must_use on the finish function

### LINKED ISSUE #77: `must_use` build()
While I like seeing the `struct FooBuilder<…>` having a `must_use` (#26) I'm not sure why the `build()` method does not have one. Some `Into` implementations could have side effects but other than that it should just create an object which hasn't been used so far. So its likely wrong when the `build()` output isn't used?

EdJoPaTo:
closes #77

The implementation for cloning the `must_use` of the original function can likely be optimized. Wanted to bring this up to have something to discuss.

Maybe the `#[must_use = …]` of the original function should be removed as it could get "expensive" with this kind of usage: `#[must_use = include_str!("…")]`. It will only be called internally, so there is not much benefit of having the `must_use` there then?

### DIFF
--- a/bon-macros/src/builder/builder_gen/input_func.rs
+++ b/bon-macros/src/builder/builder_gen/input_func.rs
@@ -326,7 +326,7 @@ impl FuncInputCtx {
             ident: finish_func_ident,
             unsafety: self.norm_func.sig.unsafety,
             asyncness: self.norm_func.sig.asyncness,
-            must_use: get_must_use_attribute(&self.norm_func.attrs),
+            must_use: get_must_use_attribute(&self.norm_func.attrs)?,
             body: Box::new(finish_func_body),
             output: self.norm_func.sig.output,
             docs: "Finishes building and performs the requested action.".to_owned(),
@@ -505,26 +505,16 @@ impl Visit<'_> for FindSelfReference {
     }
 }
 
-fn get_must_use_attribute(attrs: &[syn::Attribute]) -> Option<syn::Attribute> {
+fn get_must_use_attribute(attrs: &[syn::Attribute]) -> darling::Result<Option<syn::Attribute>> {
     let mut iter = attrs
         .iter()
-        .filter(|attr| !attr.is_doc() && matches!(attr.style, syn::AttrStyle::Outer))
-        .filter(|attr| {
-            let path = match &attr.meta {
-                syn::Meta::Path(path) => path,
-                syn::Meta::List(_) => return false,
-                syn::Meta::NameValue(name_value) => &name_value.path,
-            };
-            path.segments.len() == 1
-                && path
-                    .segments
-                    .iter()
-                    .any(|segment| segment.ident.raw_name() == "must_use")
-        });
+        .filter(|attr| attr.meta.path().is_ident("must_use"));
     let result = iter.next();
-    if iter.next().is_some() {
-        // Fishy: multiple must_use? panic here?
-        return None;
+    if let Some(second) = iter.next() {
+        bail!(
+            second,
+            "Found multiple #[must_use], but bon only works with exactly one (or less)."
+        );
     }
-    result.cloned()
+    Ok(result.cloned())
 }
--- a/bon-macros/src/builder/builder_gen/input_func.rs
+++ b/bon-macros/src/builder/builder_gen/input_func.rs
@@ -326,6 +326,7 @@ impl FuncInputCtx {
             ident: finish_func_ident,
             unsafety: self.norm_func.sig.unsafety,
             asyncness: self.norm_func.sig.asyncness,
+            must_use: get_must_use_attribute(&self.norm_func.attrs),
             body: Box::new(finish_func_body),
             output: self.norm_func.sig.output,
             docs: "Finishes building and performs the requested action.".to_owned(),
@@ -502,4 +503,28 @@ impl Visit<'_> for FindSelfReference {
             self.self_span = Some(first_segment.ident.span());
         }
     }
+}
+
+fn get_must_use_attribute(attrs: &[syn::Attribute]) -> Option<syn::Attribute> {
+    let mut iter = attrs
+        .iter()
+        .filter(|attr| !attr.is_doc() && matches!(attr.style, syn::AttrStyle::Outer))
+        .filter(|attr| {
+            let path = match &attr.meta {
+                syn::Meta::Path(path) => path,
+                syn::Meta::List(_) => return false,
+                syn::Meta::NameValue(name_value) => &name_value.path,
+            };
+            path.segments.len() == 1
+                && path
+                    .segments
+                    .iter()
+                    .any(|segment| segment.ident.raw_name() == "must_use")
+        });
+    let result = iter.next();
+    if iter.next().is_some() {
+        // Fishy: multiple must_use? panic here?
+        return None;
+    }
+    result.cloned()
 }
--- a/bon-macros/src/builder/builder_gen/input_struct.rs
+++ b/bon-macros/src/builder/builder_gen/input_struct.rs
@@ -126,6 +126,9 @@ impl StructInputCtx {
             ident: finish_func_ident,
             unsafety: None,
             asyncness: None,
+            must_use: Some(syn::parse_quote! {
+                #[must_use = "Building a struct without using it is likely a bug"]
+            }),
             body: Box::new(finish_func_body),
             output: syn::parse_quote!(-> #struct_ty),
             docs: "Finishes building and returns the requested object.".to_owned(),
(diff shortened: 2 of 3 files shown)


bon-macros/src/builder/builder_gen/input_func.rs
```diff
@@ -503,3 +504,27 @@ impl Visit<'_> for FindSelfReference {
...
+            };
+            path.segments.len() == 1
+                && path
+                    .segments
+                    .iter()
+                    .any(|segment| segment.ident.raw_name() == "must_use")
+        });
+    let result = iter.next();
+    if iter.next().is_some() {
+        // Fishy: multiple must_use? panic here?
```
Veetaha, line 526: This results in a compiler warning:
```rust
#[must_use = "first"]
#[must_use = "second"]
fn foo() {}
```
```
warning: unused attribute
 --> crates/sandbox/src/lib.rs:2:1
  |
2 | #[must_use = "second"]
  | ^^^^^^^^^^^^^^^^^^^^^^ help: remove this attribute
  |
note: attribute also specified here
 --> crates/sandbox/src/lib.rs:1:1
  |
1 | #[must_use = "first"]
  | ^^^^^^^^^^^^^^^^^^^^^
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: `#[warn(unused_attributes)]` on by default
```
I think in this case we should just return an error, with the span pointing to the second usage of the attribute e.g.:
```rust
if let Some(second_attr) = iter.next() {
    bail!(second_attr, "Found multiple #[must_use], but only one is allowed");    
}
```
This way the error message will point to the second `#[must_use]` attribute. You can test this by adding a test case to `/home/veetaha/work/bon/bon/tests/integration/ui/compile_fail/misc.rs`. Then update the `misc.stderr` snapshot with 
```
TRYBUILD=overwrite cargo t --all-features --all-targets -p bon
```
And inspect the git diff in the error messages in that stderr output.

Btw. I also recommend you to add a test that creates a struct and doesn't use it or calls a function with `must_use` without using its result into the `compile_fail/misc.rs` file. This file is used to test that the compiler is able to generate the expceted error messages and warnings.

bon-macros/src/builder/builder_gen/input_func.rs
```diff
@@ -503,3 +504,27 @@ impl Visit<'_> for FindSelfReference {
...
+            let path = match &attr.meta {
+                syn::Meta::Path(path) => path,
+                syn::Meta::List(_) => return false,
+                syn::Meta::NameValue(name_value) => &name_value.path,
+            };
+            path.segments.len() == 1
+                && path
+                    .segments
+                    .iter()
+                    .any(|segment| segment.ident.raw_name() == "must_use")
```
Veetaha, line 522: Let's follow the pattern of adding extension trait methods to `syn::Attribute` like here: https://github.com/elastio/bon/blob/741ed74217e922e3b6e0fa4480b59f5469469880/bon-macros/src/util/attrs.rs#L11-L21

Btw. a simpler way to check that `syn::Path` is equal to `must_use` is with the [`syn::Path::is_ident()`](https://docs.rs/syn/latest/syn/struct.Path.html#method.is_ident) method.

bon-macros/src/builder/builder_gen/input_func.rs
```diff
@@ -503,3 +504,27 @@ impl Visit<'_> for FindSelfReference {
         }
     }
 }
+
+fn get_must_use_attribute(attrs: &[syn::Attribute]) -> Option<syn::Attribute> {
+    let mut iter = attrs
+        .iter()
+        .filter(|attr| !attr.is_doc() && matches!(attr.style, syn::AttrStyle::Outer))
```
Veetaha, line 511: I think this entire `filter(...)` here is redundant because the filter below is a subset of the what's checked here. For example `doc` attributes definitely don't have a `must_use` path in them, and nothing but outer attributes are supported on functions.

Veetaha (commented):
> Maybe the #[must_use = …] of the original function should be removed as it could get "expensive" with this kind of usage: #[must_use = include_str!("…")].

I don't think anyone would use `include_str!()` with `#[must_use]`. Even if they do, the message of `#[must_use]` is not something where people would put gigabytes of text. There will probably be a line of text that fits into a conventional 80-120 chracter line limit.

In general, the proc macros here operate on a rather limited set of data (just a single function or struct or an impl block), which are not that expensive to clone. There are a lot of clones already happening in `bon-macros`, for example, we clone the entire function/struct right at the start of the macro, e.g. here:
https://github.com/elastio/bon/blob/741ed74217e922e3b6e0fa4480b59f5469469880/bon-macros/src/builder/item_func.rs#L8-L11

This is done to have both the original (unmodified) version of the syntax, and the normalized version of the syntax (where the syntax sugar of `impl Trait`, anonymous and elided lifetimes and `Self` references is removed.

So cloning a mere `#[must_use]` attribute isn't a big deal.

---

> It will only be called internally, so there is not much benefit of having the must_use there then?

The `#[must_use]` must be retained on the positional function. The rule of thumb is to avoid mutating the original code written by the user as much as possible (which may lead to some errors in original syntax being removed and unnoticed by the compiler, for example). But in this case, the original positional function can also be exposed, and the users can call it directly if they use the [`expose_positional_fn`](https://elastio.github.io/bon/reference/builder#expose-positional-fn) attribute.

bon-macros/src/builder/builder_gen/input_func.rs
Veetaha: This makes it for a fun reddit/twitter post:
- [Reddit](https://www.reddit.com/r/rust/comments/1f4vvgy/media_fun_fact_you_can_write_functions_docs_in/)
- [Twitter](https://x.com/veetaha/status/1829516618386588060)

EdJoPaTo: Without the filter it might also find `must_use` defined as inner attribute? It's still horrible… :D

Veetaha: Yeah, we can either filter inner attributes, or produce a compile error if someone uses an inner doc or must_use attribute (they deserve it 👀 )

EdJoPaTo: Its horrible, yes, but we probably don't need to filter at all. syn seems to correctly parse them, so why not detect either case and just use it? Its horrible code, but we have one filter less, and it still works. Less complexity on the macro side.

(They still might deserve it :axe:)

Veetaha: Yeah, if it doesn't cost us anything in terms of code complexity, let's do that

(But they don't deserve this 💯 )

EdJoPaTo: bon adds outer attributes so rust is unhappy about the inner attributes being used. On stable it doesn't work, on nightly its annoyed. No need to create any way around it, it doesn't work in the first place. :D

bon-macros/src/builder/builder_gen/input_func.rs
```diff
@@ -503,3 +504,17 @@ impl Visit<'_> for FindSelfReference {
         }
     }
 }
+
+fn get_must_use_attribute(attrs: &[syn::Attribute]) -> darling::Result<Option<syn::Attribute>> {
```
Veetaha, line 508: The `Result` type of the crate is already in scope via `crate::util::prelude::*` import, no need to qualify it.
```suggestion
fn get_must_use_attribute(attrs: &[syn::Attribute]) -> Result<Option<syn::Attribute>> {
```

EdJoPaTo: Ugh, Result ≠ Result… I dislike prelude, wildcard imports and shadowing the std/core prelude as it quickly gets confusing. :D

EdJoPaTo:
Just as a heads up: I will likely add test cases no earlier than Sunday, more likely Monday.

Veetaha (approved):

Veetaha:
I decided to merge the PR in the current state, and add tests from my side, because I'm rewriting the builder signatures to reduce the type-checking time required for the generated code (https://github.com/elastio/bon/pull/84) and it touches a lot of code, including one touched in this PR. So, to avoid future merge conflicts, I'm merging this now.

Thank you for the PR though!

Veetaha:
This and other changes were released as part of the `2.1` release:

- [Blog post](https://elastio.github.io/bon/blog/bon-builder-v2-1-release)
- [Reddit post](https://www.reddit.com/r/rust/comments/1f6d7vr/media_nextgen_builder_macro_bon_21_release/)
- [X (Twitter)](https://x.com/veetaha/status/1830221731677839637)

EdJoPaTo:
Thank you for guiding me with the PR! Was nice to tinker with proc macros for the first time!

Issue #81, opened 2024-08-29
Set an msrv requirement and test it on CI


Veetaha:
The current code is tested with `1.80.1` on CI, but I tried to keep it compatible with older versions manually. 
We need to set the `required-rust-version` in Cargo.toml and the MSRV version to the tests on CI. CI should have matrix jobs for each check that validate the code on both MSRV and latest stable

EdJoPaTo:
Personally I would try to get 1.63 (or earlier) as MSRV in order to support current Debian bullseye. That would allow for tools depending on bon to be packages natively for Debian Bullseye.

Also, `-Zmsrv-policy` (part of the upcoming `resolver = "3"` / edition 2024) should help with that. See https://github.com/rust-lang/cargo/issues/13873

The following graphic contains the Rust version in currently supported (no out of support) distros known to repology.org:

[![Packaging status](https://repology.org/badge/vertical-allrepos/rust.svg?columns=2&exclude_unsupported=1)](https://repology.org/project/rust/versions)

Veetaha:
Thank you for the insight! [`derive-builder`](https://github.com/colin-kiegel/rust-derive-builder) supports `1.56`, although it depends on `syn 2.0` that supports `1.61+`. I believe it's better to follow `syn` in this regard, since any change to `syn` may accidentally break its current untested compatibility with `rustc 1.56`.

Veetaha:
Correction `syn 2.0.56` supports rust version `1.56`, but `syn 2.0.57` bumped that requirement. I think to be able to replace `derive-builder` we need to do the same then. Not sure how hard it'll be and if it will be hard I'm fine to use `1.61` as MSRV to follow the cutting edge of `syn`.

Veetaha:
> Personally I think it's fine to bump MSRV when the `rust-version` on the published versions on crates.io is correct

What do you mean by "when the `rust-version` is correct"? If it's incorrect, then it's a bujg

Veetaha:
I've set MSRV to 1.70.0 for now. The problem with making it lower is that crates we use for testing (such as trybuild, for example), don't compile on versions lower than that. I may consider lowering this requirement in the future if a compelling use case emerges.

Veetaha:
I created a followup issue to keep lowering the MSRV (#102). This first iteration is just a good enough first attempt at lowering it.

Veetaha:
[...]

PR #101, opened 2024-09-03
Set MSRV to 1.70.0. Fix lints triggered by generated code

Veetaha:
Closes https://github.com/elastio/bon/issues/81


bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -233,6 +237,15 @@ impl BuilderGenCtx {
...
             #(#docs)*
             #[inline(always)]
+            #[allow(
+                // This is intentional. We want the builder syntax to compile away
+                clippy::inline_always,
+                // We normalize `Self` references intentionally to simplify code generation
+                clippy::use_self,
+                // Let's keep it as non-const for now to avoid restricting ourselfves to only
+                // const operations.
+                clippy::missing_const_for_fn
```
EdJoPaTo, line 247: I'm curious whether `#[inline(always)]` will result in a kind of const operation anyway due to optimizations :thinking:

Having a const function might be better so maybe revisit this in the future?

anelson (approved):

e2e-tests/src/lib.rs
```diff
@@ -36,6 +36,7 @@ impl Counter {
         #[builder(default)]
         initial: usize,
     ) -> Self {
+        eprintln!("Non-const");
```
EdJoPaTo, line 39: Is this to prevent `clippy::missing_const_for_fn`? Do that explicitly rather than some magic eprintln without context?

Cargo.toml
```diff
@@ -132,6 +135,9 @@ verbose_bit_mask                  = "warn"
 wildcard_dependencies             = "warn"
 zero_sized_map_values             = "warn"
 
+nursery  = { level = "warn", priority = -2 }
+pedantic = { level = "warn", priority = -1 }
```
EdJoPaTo, line 139: Why different priorities? Having them both as -1 is enough to override them.

bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -363,6 +376,16 @@ impl BuilderGenCtx {
...
             #allows
+            #[allow(
+                // We use `__private` prefix for all fields intentionally to hide them
+                clippy::struct_field_names,
+
+                // This lint doesn't emerge until you manually expand the macro. Just
+                // because `bon` developers need to expand the macros a lot it makes
+                // sense to just silence it to avoid some noise. This lint is triggered
+                // by the big PhantomData type generated by the macro
+                clippy::type_complexity
```
EdJoPaTo, line 387: Allow it on the PhantomData then directly?

bon/tests/integration/main.rs
```diff
@@ -24,13 +30,14 @@ use expect_test::Expect;
 const COMMON_SCREEN_CHARS_WIDTH: usize = 60;
 
 #[track_caller]
+#[allow(clippy::needless_pass_by_value)]
```
EdJoPaTo, line 33: allowed at the top and here

e2e-tests/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "e2e-tests"
 
-publish = false
+# publish = false
```
EdJoPaTo, line 4: Why comment out? Either remove or keep there?

EdJoPaTo (commented):
This PR is about adding more lints, specifying the MSRV is just a byproduct? The goal of this PR feels a bit random as all kinds of things are happening. Maybe split it up?

Cargo.toml
```diff
@@ -6,6 +6,9 @@ resolver = "2"
 edition = "2021"
 version = "2.1.0"
 
+# MSRV. If you need a lower MSRV, please create an issue and we'll consider lowering it.
+rust-version = "1.70.0"
```
EdJoPaTo, line 10: This also requires `rust-version = { workspace = true }`, doesnt it?

Veetaha: Good catch! Fixed it, thanks

e2e-tests/Cargo.toml
Veetaha: That's some debugging remnant, thank you for noticing this, I'll remove

bon-macros/src/builder/builder_gen/mod.rs
Veetaha: I'm 99% sure this will be come a `const fn` in the future. I'm just conservative about this, and I'd like to keep the posibility to use non-const operators here for `bon`, since I believe there are no uses cases for `const` builders, and besides just marking this fn `const` isn't enough. We need to have `IntoSet` trait const as well

e2e-tests/src/lib.rs
Veetaha: I use this `eprinln!()` to avoid adding a global  `allow`. This way if `bon` generates code that violates this lint (which it did before) then clippy will catch this.

Cargo.toml
Veetaha: Priorities are used not because we override lints from these lint groups but just to order them in order from the less noisy to the more noisy in the output. I'll add a code comment

bon-macros/src/builder/builder_gen/mod.rs
Veetaha: I'll keep it at struct level. This whole struct is generated and in case some other field starts having "big type complexity" this lint will cover it as well. So it's just for the future

bon/tests/integration/main.rs
Veetaha: Makes sense, removed

Veetaha:
> This PR is about adding more lints, specifying the MSRV is just a byproduct?

This is both, because it tackles one of the real use cases where MSRV  `<=1.75.0` is required plus fixing the lints from the generated code. I don't think it makes sense to split the PR, especially because the work is already done and I don't see any sense in the ceremony of splitting it

Veetaha:
Note that `1.70.0` is not the real lowest MSRV possible here. It's just the one I could ensure quickly to make a patch release and thus unblock some of the users of `bon` that require the MSRV `1.75.0`.

I created an issue https://github.com/elastio/bon/issues/102 to keep the MSRV even lower

PR #105, opened 2024-09-04
[Docs] Update broken links

kageiit:
Fixed links to the docs around from/into and reasons to avoid/use

Thanks for the great work on bon!

Also added a step using https://github.com/UmbrellaDocs/action-linkspector to detect issue moving forward on PRs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,6 +196,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: crate-ci/typos@v1.22.3
+  
+  # Check for broken links in docs
+  check-links:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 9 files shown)


Veetaha:
Thank you for the PR!

I thought Vitepress already detects broken anchors in URLs because it definitely has an inbuilt link checker, but turns out it's not so advanced yet https://github.com/vuejs/vitepress/issues/646.

I've run the CI and turns out linkspector doesn't detect the broken links. A kind person shared their custom script that detects broken links in https://github.com/vuejs/vitepress/issues/646#issuecomment-1722916907. I'll use that instead.

Veetaha (approved):
Great thanks to @0x009922 for saving our docs from broken anchors ❤️ ,  and thanks @kageiit for the PR as well 🐱, I'll update the website after the merge

Issue #112, opened 2024-09-08
chore: updated readme with marketing targeted to functional programmers


dzmitry-lahoda:
> bon is a Rust crate for generating compile-time-checked builders for functions and structs.

builder are fine, but for many they bring negative connotation. 

but in languages, specifically F#, but guess many others, there are notions of `partial application` and `named parameters`. 

so basically bon is efficient way to implement these (there are failed attempts in Rust to do so).

so can improve search ability and appreciation of bon by:

 > bon is a Rust crate for generating compile-time-checked builders for functions, methods and structs,
>  idiomatic partial application, optional and named parameters for functions and methods

Veetaha:
I'm fine with this change to Readme. I just tried to keep it simple for those beginner Rust developers, but I see people often asking questions like "why is it useful? does it support optional parameters", so I guess this change is for the better. 

Although, I have to note that partial application in the current state of `bon` is useful only in the patterns of conditional building at most.

dzmitry-lahoda:
> I have to note that partial application in the current state of bon is useful only in the patterns of conditional building at most. The type signature of the partial builder is not stable yet, so partial application is useful only in a local scope.

Local scope is fine.

My case is proptest data builder or  setup for fuzzying or some ddos bot builder.

There whole frameworks build on that, like zio or https://effect.website/ . So it may still be useful.

For production case may appear to be HTTP, when need to negotiate content type or route depending on headers.

If there are chances `bon` builder will get public type at any point, than README is fine and future proof.

Veetaha:
> ddos bot builder

What is the ddos target?

dzmitry-lahoda:
> What is the target of ddos?

In order to try to stress/load test services, need to produce a lot of values which penetrate it functional deeply. For this need to use some input data builder to build reasonable inputs.

PR #114, opened 2024-09-08
chore(docs):  readme which describer full feature set for #112

dzmitry-lahoda:
for https://github.com/elastio/bon/issues/112 

used `idiomatic` keyword because there are attempts to do partial application and named parameters in haskell/scala style which all non idiomatic and not nice looking Rust people expect.

so bon approach seems like nice fit

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,14 @@
     /></a>
 </p>
 
-`bon` is a Rust crate for generating compile-time-checked builders for functions and structs.
+
+`bon` is a Rust crate for generating compile-time-checked builders for functions and structs, supporting partial struct building and idiomatic partial application with optional and named parameters for functions and methods.
 
 Visit the [guide for a complete overview of the crate](https://elastio.github.io/bon/guide/overview).
 
 ## Quick examples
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,13 +27,14 @@
...
-`bon` is a Rust crate for generating compile-time-checked builders for functions and structs.
+
+`bon` is a Rust crate for generating compile-time-checked builders for functions and structs, supporting partial struct building and idiomatic partial application with optional and named parameters for functions and methods.
 
 Visit the [guide for a complete overview of the crate](https://elastio.github.io/bon/guide/overview).
 
 ## Quick examples
 
-`bon` can turn a function with positional parameters into a function with "named" parameters via a builder. It's as easy as placing the `#[builder]` attribute on top of it.
+`bon` turns a function with positional parameters into a function with named parameters just by placing the `#[builder]` attribute on top of it.
```
Veetaha, line 37: `just by` placing does sound right as if `bon` is the one who palces the `#[builder]` attribute when in fact, it's the developer who does this. Anyway, I'll make my corrections to the readme in #113

Veetaha (approved):
Thank you for the PR! I'm going to substantially update the README in https://github.com/elastio/bon/pull/113 before the `2.2` release.

Issue #111, opened 2024-09-08
feat: allow to access already set fields within pattern of partial shared builder 


dzmitry-lahoda:
```rust
use bon::builder;

#[builder(on(String, into))]
struct User {
    name: String,
    tags: Vec<String>,
    alias: Option<String>,
    description: Option<String>,
}

// Common part is here
let user = User::builder()
    .name("Bon")
    .tags(vec!["confectioner".to_owned()]);

// Differences are in each branch
let user = if user.GET_name().as_str() == "Bonny" {
    user
        .alias("Good girl")
        .description("Knows mathematics 🐱")
        .build()
} else {
    user
        .description("Skipped math classes 😿")
        .build()
};
```

## real world usage

`proptest` allows to chain into hierarchy generated properties via `prop_flat_map` (with optimized shrinking retained). 

so each `prop_flat_map` in chain depends on data to be generated from upper layer, uses that data to constraint its out put.

`prop_flat_map(|prev| prev, prev+1 ).prop_flat_map(|(prev_0, prev_1|, ..., 0..prev_1).` 

so final test function gets `do_test(prev_0, prev_1, prev_2)`. 

so I thin of using builder on each stage and path only partial builder, instead of many parameters. but need access to already set fields. in the end I do call build and produce final data struct.

example without builder:
```rust
    runner
        .run(
            &(
                1..1_000_000u64,
                0..(7u8 + 1),
                0..(6u8 + 1),
                0..(6u8 + 1),
                Margins::valid_margins_strategy(),
            )
                .prop_flat_map(
                    |(
                        market_max_balance_in_units,
                        base_token_decimals,
                        quote_token_decimals,
                        market_price_decimals,
                        market_margins,
                    )| {
                        (
                            Just(market_max_balance_in_units),
                            Just(base_token_decimals),
                            Just(quote_token_decimals),
                            Just(market_price_decimals),
                            1u64..(market_max_balance_in_units + 1),
                            0..(base_token_decimals + 1),
                            1..(Decimals::from_repr(market_price_decimals)
                                .add_u8(2)
                                .scale_u64()
                                + 1), // price_mantissa
                            Just(market_margins),
                            // TODO: variate markets and token counts
                            // TODO: variate Weigth and Margins - make sure that balance of user is inverse to this
                            // TODO: variate many makers vs one taker (ensure full fill always)
                            // TODO: pepr market
                            // TODO: variate limit via size/price/quote (use inverse formuals to set limit correct)
[...]

PR #118, opened 2024-09-08
feat: getter attribute for #111

dzmitry-lahoda:
- for #111
- minimal possible impl but can be grown to getset crate level of features
- no docs, hidden as unstable
- added to setters, because in that place members are iterated; renamed plural setter builder to singular. getter is small, did not saw reason to create new mod

### DIFF
--- a/bon-macros/src/builder/builder_gen/member.rs
+++ b/bon-macros/src/builder/builder_gen/member.rs
@@ -109,6 +109,9 @@ pub(crate) struct MemberParams {
 
     /// Rename the name exposed in the builder API.
     pub(crate) name: Option<syn::Ident>,
+
+    /// Whether to generate a public reference getter method for this member.
+    pub(crate) getter: darling::util::Flag,
 }
 
 impl MemberParams {
@@ -118,6 +121,7 @@ impl MemberParams {
             into,
             default,
             name,
+            getter,
         } = self
         {
             match origin {
@@ -152,6 +156,13 @@ impl MemberParams {
                     &span,
                     "`skip` attribute can't be specified with other attributes like `{attr_name}` \
                     because there will be no setter generated for this member to configure{default_hint}",
+                );
+            }
+
+            if !cfg!(feature = "getter") && getter.is_present() {
+                bail!(
+                    &getter.span(),
+                    "`getter` attribute is only available when the `getter` feature is enabled",
                 );
             }
         }
@@ -244,8 +255,12 @@ impl RegularMember {
         Ok(verdict_from_override || verdict_from_defaults)
     }
 
-    pub(crate) fn setter_method_core_name(&self) -> &syn::Ident {
+    pub(crate) fn accessor_method_core_name(&self) -> &syn::Ident {
         self.params.name.as_ref().unwrap_or(&self.norm_ident)
+    }
+
+    pub(crate) fn param_getter(&self) -> Option<()> {
+        self.params.getter.is_present().then(|| ())
     }
 }
 
--- a/bon-macros/src/builder/builder_gen/mod.rs
+++ b/bon-macros/src/builder/builder_gen/mod.rs
@@ -5,6 +5,8 @@ mod setter_methods;
 pub(crate) mod input_func;
 pub(crate) mod input_struct;
 
+use core::iter::once;
+
 use member::*;
 
 use super::params::{BuilderDerives, ConditionalParams};
@@ -197,7 +199,7 @@ impl BuilderGenCtx {
 
     fn builder_impl(&self) -> Result<TokenStream2> {
         let finish_method = self.finish_method()?;
-        let (setter_methods, items_for_rustdoc) = self.setter_methods()?;
+        let (accessor_methods, items_for_rustdoc) = self.accessor_methods()?;
 
         let generics_decl = &self.generics.decl_without_defaults;
         let generic_args = &self.generics.args;
@@ -239,7 +241,7 @@ impl BuilderGenCtx {
             #where_clause
             {
                 #finish_method
-                #setter_methods
+                #accessor_methods
             }
         })
     }
@@ -562,7 +564,7 @@ impl BuilderGenCtx {
         quote::format_ident!(
             "{}__{}",
             self.builder_ident.raw_name(),
-            member.setter_method_core_name()
+            member.accessor_method_core_name()
         )
     }
 
@@ -627,7 +629,7 @@ impl BuilderGenCtx {
         })
     }
 
-    fn setter_methods(&self) -> Result<(TokenStream2, TokenStream2)> {
+    fn accessor_methods(&self) -> Result<(TokenStream2, TokenStream2)> {
         let generics_decl = &self.generics.decl_without_defaults;
         let generic_args = &self.generics.args;
         let builder_ident = &self.builder_ident;
@@ -650,7 +652,7 @@ impl BuilderGenCtx {
             }
         });
 
-        let setters = self
+        let accessors = self
             .regular_members()
             .map(|member| {
                 let state_types = self.regular_members().map(|other_member| {
@@ -676,15 +678,17 @@ impl BuilderGenCtx {
                     doc_false: next_state.clone(),
                 };
 
-                let setter_methods =
+                let setter_method =
                     MemberSettersCtx::new(self, member, return_type).setter_methods()?;
+
+                let maybe_getter = self.getter_method(member);
 
                 let next_state = quote!(type #member_pascal = #next_state;);
 
-                Ok((setter_methods, next_state))
+                Ok((setter_method, next_state, maybe_getter))
             })
             .collect::<Result<Vec<_>>>()?;
-        let next_states_defs = setters.iter().map(|(_, next_state)| next_state);
+        let next_states_defs = accessors.iter().map(|(_, next_state, _)| next_state);
 
         let items_for_rustdoc = quote! {
             // This item is under `cfg(doc)` because it's used only to make the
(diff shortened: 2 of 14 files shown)


dzmitry-lahoda:
cc @Veetaha

bon-macros/src/builder/builder_gen/member.rs
```diff
@@ -109,6 +109,9 @@ pub(crate) struct MemberParams {
 
     /// Rename the name exposed in the builder API.
     pub(crate) name: Option<syn::Ident>,
+
+    /// Whether to generate a public reference getter method for this member.
+    pub(crate) get: darling::util::Flag,
```
Veetaha, line 114: I'm re-thinking that maybe a better name for this could be `#[builder(getter)]`. I think this way it's easier for people not familiar with `bon` to understand what the attribute does.

The options for the getter (if and when we add them) will be inside of parentheses like:

```rust
#[builder(
    getter(
        vis = "pub(crate)",

        name = override_name,
        
        // Return a `Copy` of the value
        copy,
        // ...
    )
)]
```

bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -197,7 +197,7 @@ impl BuilderGenCtx {
 
     fn builder_impl(&self) -> Result<TokenStream2> {
         let finish_method = self.finish_method()?;
-        let (setter_methods, items_for_rustdoc) = self.setter_methods()?;
+        let (setget_methods, items_for_rustdoc) = self.setget_methods()?;
```
Veetaha, line 200: I think the name `accessor` fits better here as the generalization of getters and setters and this name will also scale. For example, if we add a way to mutate the fields like in [`std::process::Command`](https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.arg) builder where you can invoke `.arg()` multiple time and it pushes a new argument every time the method is called. In [`typed-builder`](https://docs.rs/typed-builder/latest/typed_builder/derive.TypedBuilder.html)'s terminology such methods are called mutators (search their docs page by the word `mutator`).

I'm also planning to have some way to define mutators in `bon`, I'd like to avoid the syntax of defining an `impl` block inside of the attribute that's used inside of `typed-builder`. A more nature way to express that API in that I'm thinking to have in the future is by letting the user write an `impl` block agains the builder type directly.

```suggestion
        let (accessor_methods, items_for_rustdoc) = self.accessor_methods()?;
```

bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -715,12 +717,59 @@ impl BuilderGenCtx {
             }
         };
 
-        let setter_methods = setters
+        let setget_methods = setgets
             .into_iter()
-            .map(|(setter_methods, _)| setter_methods)
+            .map(|(setter_method, _, maybe_getter_method)| {
+                [setter_method, maybe_getter_method.unwrap_or_default()].concat()
+            })
```
Veetaha, line 724: Nested concat can be avoided like this:

```suggestion
            .flat_map(|(setter_method, _, maybe_getter_method)| {
                [setter_method].into_iter().chain(maybe_getter_method)
            })
```
This is thanks to the fact that `Option` implements `IntoIterator`

bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -715,12 +717,59 @@ impl BuilderGenCtx {
...
             .concat();
 
-        Ok((setter_methods, items_for_rustdoc))
+        Ok((setget_methods, items_for_rustdoc))
+    }
+
+    fn getter_method(&self, member: &RegularMember) -> Option<TokenStream2> {
+        member.param_getter().map(|()| {
```
Veetaha, line 731: Just use the `?` operator and de-nest the rest of the code
```suggestion
        member.param_getter()?
```

bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -715,12 +717,59 @@ impl BuilderGenCtx {
...
+            // for comments on how this getter formed, please look to setter
```
Veetaha, line 732: The `setter_methods.rs` file is quite a big one, and if you wanted to comment on something I'd recommend just copying the comment here or using some shared method between setters and getters

bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -715,12 +717,59 @@ impl BuilderGenCtx {
...
+            let member_type = member.norm_ty.as_ref();
+            let member_index = &member.index;
+            let member_state_type =  &member.generic_var_ident;
+            let vis = &self.vis;
+            let member_label = self.members_label(member);
+            let getter_method_name = syn::Ident::new(
+                &format!("get_{}", member.orig_ident),
+                member.orig_ident.span(),
+            );
```
Veetaha, line 741: Instead of `member.orig_ident`, this should use the `member.setter_method_core_name()`, which take into the account the name override with `#[builder(name = newname)]`. I think it makes sense to reuse that override for the getter as well. Note that we'll need to rename the `setter_method_core_name` to `accessor_method_core_name` to make it generic and shared between setters and getters.

bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -715,12 +717,59 @@ impl BuilderGenCtx {
...
+
+            if let Some(inner_type) = member.as_optional_norm_ty() {
+                quote! {
+                    /// Unstable
```
Veetaha, line 745: Let's write some actual documentation for this method. For example:

```rust
let docs = format!("Returns the value of `{}`", member.accessor_method_core_name());
```

shell.nix
```diff
@@ -0,0 +1,15 @@
+# just qick way to get exact rust used with all components and targets on mac and linux
```
Veetaha, line 1: I'd like to avoid commiting tool-specific configs like this into the repo. I think most of the developers just use `rustup` to install rust, and I don't want to make it more complex than that in this repo, given that it's just a rust crate that doesn't depend on any OS-level libraries and binaries.

Veetaha (commented):
In addition to my comments I propose to create a cargo feature flag called `"unstable"` then inside of this `validate` method in `Member`:
https://github.com/elastio/bon/blob/c0dd9c9ee4184e27b93e88d9ffec5863b7888d25/bon-macros/src/builder/builder_gen/member.rs#L115

Check that `cfg!(feature = "unstable")` returns true if a `getter` was requested and if not, return an error that tells the user that they are using an unstable feature that may break between patch releases and they should confirm they are alright with using it by adding the `"unstable"` feature flag to the `bon` dependency.

bon-macros/src/builder/builder_gen/setter_methods.rs
```diff
@@ -35,7 +35,7 @@ impl<'a> MemberSettersCtx<'a> {
         }
     }
 
-    pub(crate) fn setter_methods(&self) -> Result<TokenStream2> {
+    pub(crate) fn try_setter_method(&self) -> Result<TokenStream2> {
```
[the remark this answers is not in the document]
dzmitry-lahoda: why it called `setter_methodS`? it is single setter_method.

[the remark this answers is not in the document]
Veetaha: This method returns two settter methods for optional parameters as shown on the snippet above, so the name of the method is plural

bon-macros/src/builder/builder_gen/member.rs
dzmitry-lahoda: used getset as reference example

Veetaha: Yeah, in `getset` the `get` attribute is well-known because it's their primary API, but in `#[builder]` it's going to be used rarely, so a bit more verbose but descriptive name is better IMHO

bon-macros/src/builder/builder_gen/setter_methods.rs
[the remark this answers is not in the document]
dzmitry-lahoda: ok, it returns one method body, but ok. will return back.

bon/tests/integration/builder/attr_get.rs
```diff
@@ -0,0 +1,39 @@
+// Please note that most parts of field variability in naming tested by getters
+#[cfg(feature = "alloc")]
+#[test]
+fn compiles_and_works() {
+    use crate::alloc::string::ToString;
+    use crate::prelude::*;
+    use alloc::boxed::Box;
+    use core::pin::Pin;
+    use core::ptr::NonNull;
```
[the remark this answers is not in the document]
dzmitry-lahoda: will not use alloc. avoid these changes

bon/tests/integration/builder/attr_get.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    let abcde = Abcde::builder().aaa(1);
+    assert_eq!(abcde.get_a(), &1);
+
+    let abcde = abcde.maybe_d(Some(vec![1, 2, 3]));
+    assert_eq!(abcde.get_d(), Some(&vec![1, 2, 3]));
+
+    let abcde = abcde.b("hello".to_string());
+    assert_eq!(abcde.get_b(), &"hello".to_string());
```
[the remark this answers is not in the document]
dzmitry-lahoda: will not use alloc. avoid these changes

bon/tests/integration/builder/attr_get.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    // just doing some crazy type reference
+    let mut stack = 42u8;
+    #[allow(unsafe_code)]
+    let pin = unsafe { Pin::new(Box::new(NonNull::new_unchecked(&mut stack))) };
+    let abcde = abcde.e(&pin);
+    let _e: &Pin<Box<NonNull<u8>>> = abcde.get_e();
```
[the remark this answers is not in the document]
dzmitry-lahoda: will not use alloc. avoid these changes

bon-macros/src/builder/builder_gen/mod.rs
dzmitry-lahoda: ok,

btw

```
    // Mutator has access to `x` additionally.
    #[mutator(requires = [x])]
    fn x_into_b(&mut self) {
        self.b.push(self.x)
    }
```

that is terrible, if i am going to rename `x`, will need to change macro? and x already is used in code, so seems they instead of doing some more automatic thing, went too way manual, so why to use builder, better write manual

bon-macros/src/builder/builder_gen/mod.rs
dzmitry-lahoda: will remove comments

dzmitry-lahoda:
too many conflicts, hard to rebase.

Issue #102, opened 2024-09-03
Use the lowest possible MSRV 


Veetaha:
The PR https://github.com/elastio/bon/pull/101 set the initial MSRV to 1.70.0. The task here is basically the followup for https://github.com/elastio/bon/issues/81. We need to try to adapt the code to compile on the lowest possible MSRV. The desirable target is 1.56, which is the MSRV of `derive_builder`.

PR #120, opened 2024-09-09
Lower MSRV to 1.59.0

Veetaha:
Closes https://github.com/elastio/bon/issues/102

I managed to bring down MSRV to 1.56.0, but then I found a blocking compile error that made me to bump MSRV to 1.59.0, which I think is good enough though. I left an explanation in the comment here:
https://github.com/elastio/bon/blob/99c66f51651b8d9a0e32fd93e1ac2a9edf427927/bon/Cargo.toml#L21-L28

@EdJoPaTo I suppose you are anyway happy with this MSRV, right?


Veetaha:
cc @dzmitry-lahoda beware of this change. There may be some usages of `bool::then_some` that you should replace with `bool::then` in your PR, for example

anelson (approved):

.github/workflows/ci.yml
```diff
@@ -98,33 +96,15 @@ jobs:
       - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           components: clippy
-          toolchain: 1.70.0
+          toolchain: 1.59.0
```
dzmitry-lahoda, line 99: 1.7 years old. so can use only rust 1.59 in code?

dzmitry-lahoda: https://github.com/elastio/bon/blob/master/rust-toolchain contains 1.81 . so only CI can check if code is right, not local cargo check?

Veetaha: Yep, we need to support older compiler versions to let other libraries that support them use `bon`

dzmitry-lahoda: oh, `step echo '1.59.0' > rust-toolchain`

dzmitry-lahoda: seems very sloppy?

Veetaha: It is, but I think it's good enough

Veetaha: > [master/rust-toolchain](https://github.com/elastio/bon/blob/master/rust-toolchain?rgh-link-date=2024-09-10T00%3A18%3A29Z) contains 1.81 . so only CI can check if code is right, not local cargo check?

Yes, because the main focus should be on the current version of rust toolchain. For example, `#[diagnostic::on_unimplemented]` isn't available in rust `1.59.0`, so that older version of Rust will get worse compile error messages. But we don't need to work in that environment, we just need some check on CI that our code is compatible with that older version.

Besides there is a [`clippy::incompatible_msrv`](https://rust-lang.github.io/rust-clippy/master/index.html#/incompatible_msrv) lint that validates that the code doesn't use APIs not stable in 1.59.0

Veetaha: If you want to check for MSRV locally, you can just manually run `./scripts/test-msrv.sh`. It will run the test in a temp directory

Issue #24, opened 2024-07-30
Supply required fields as args to `builder()`


ThomasAlban:
I'd love to have the option to supply required fields of a struct as args to the first method call (basically deriving a `new` method), leaving optional ones as separate methods you would then be able to call. This would also mean you should't have to call `.build()` at the end of the chain, because the data structure is valid 'from the get-go'. This would be great, as currently I use `derive-new` for when I have a simple struct and can't be bothered to write the `new` function, but having this built into bon would be much nicer as setters for optional methods would be derived too.

### A note for the community from the maintainers

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the maintainers with prioritizing it. You may add a comment describing your real use case related to this issue for us to better understand the problem domain.

Veetaha:
I didn't think of having a macro to derive the `new()` method in `bon`, but it probably fits the domain. However, I suppose it's mostly a solved problem today. As you mentioned `derive-new` works fine for this use case. If you'd like to have setters generated for optional fields, then there are proc macros that do that as well. For example, [`getset`](https://docs.rs/getset/latest/getset/) is a cool one, although it hasn't been updated for 3 years.

ThomasAlban:
Yeah, I agree there have been solutions but I think bon could do it much better - for example, `derive-new` doesn't support `impl Into` (or `impl IntoIterator`) for args. Also, I probably used the wrong word for 'setters', what I really meant was functions that consume self, set a value to something, and return self again, so you can chain them. Here's an example of how everything I'm talking about could work:
```rust
#[bon::new]
struct ExampleStruct {
    #[bon(from-iter)]
    field: HashSet<u32>,
    optional: Option<usize>,
    
}

fn main() {
    let example_struct = Example::new([0, 1, 2]).optional(3);
}
```
I would definitely prefer to use a feature like this built into bon, as then it could use the same 'ecosystem' of features such as `impl IntoIteratoe` and `impl Into`

Veetaha:
Note that this has the problem of allowing you to set the value for the same optional field more than once:

```rust
Example::new([0, 1, 2]).optional(3).optional(4);
```

[...]

PR #125, opened 2024-09-10
Add support for positional params in `start_fn` and `finish_fn`

Veetaha:
Partially addresses https://github.com/elastio/bon/issues/24.

It's now possible to add `#[builder(start_fn/finish_fn)]` to the members to mark them as positional arguments to the `start_fn/finish_fn` respectively.

The relative order of members with `#[builder(start_fn/finish_fn)]` matters. Such that all members with `start_fn` will go in the same order as they are declared relative to each other in the start function, and all members with `finish_fn` will go in the same order as they are declared relative to each other in the finishing method.

For anyone willing to test this feature right from the oven, use this in `Cargo.toml`:
```toml
bon = { git = "https://github.com/elastio/bon", branch = "feat/positional-only-members" }
```

Here are some tests that showcase the API:

https://github.com/elastio/bon/blob/be3c9823e4181cfd7db2fde82f04f868a89a687b/bon/tests/integration/builder/positional_members.rs#L19-L145

This is still a draft that I'm planning to finish by the end of the week. Remaining work:
- [x] Add start fn args into builder's debug output and Clone derive
- [x] Enforce the order: start_fn params, finish_fn params, and then named/skipped params in any order (this way the order of evaluation of skip/default expressions and the variables available in their scope is consistent with the order of declaration)
- [x] Add more exhaustive tests, update existing ones, validate edge cases
- [x] Update the API reference documentation on the website
- [x] Add a new page to the Guide documentation on the website
- [x] (Maybe post-merge) start a blog post for the next minor release
- [ ] Self-review and post-refactoring


Veetaha:
cc @EdJoPaTo, @dzmitry-lahoda, and @ThomasAlban. I'd be glad to hear your feedback around this design, maybe you'd have any ideas for some better syntax to express this.

ThomasAlban:
First of all, thank you very much for all your work on this crate!
I definitely prefer the
```rust
#[builder(start_fn)]
#[builder(finish_fn)]
```
over what is currently available. However, supplying fields as args to finish_fn is something I can't see myself ever using, I was more hoping for a single toggle at the top of the struct or fn which toggles every required field to be an arg to the start function. I was also hoping this would mean there would be no need for the finish_fn, further reducing the verbosity of the API. 
I realise sacrificing the finish_fn also removes the ability to check you haven't called any optional functions multiple times, but I don't think this matters as logically it makes sense that the final call is the value that is used anyway.
Let me know if there are any reasons why supplying args to finish_fn would be useful as I can't see any!

Veetaha:
>  I was more hoping for a single toggle at the top of the struct or fn which toggles every required field to be an arg to the start function.

Yeah, that's why this PR "partially addresses #24" (as written at the top of the PR description)

> Let me know if there are any reasons why supplying args to finish_fn would be useful as I can't see any!

I do think it's useful. For example an API for sending messages from one person to another

```rust
#[derive(bon::Builder)]
#[builder(start_fn = from, finish_fn = to, on(String, into))]
struct Message {
    #[builder(pos = start_fn)]
    sender: String,

    body: String,
    is_urgent: bool,

    #[builder(pos = finish_fn)]
    recipient: String,
}

let message = Message::from("Blackjack")
    .body("Hello, Bon!")
    .is_urgent(true)
    .to("Bon");
```

Another use case is the API similar to [sqlx](https://docs.rs/sqlx/latest/sqlx/macro.query.html):

```rust
let account = sqlx::query!("select (1) as id, 'Herp Derpinson' as name")
    .fetch_one(&mut conn)
    .await?;
```
Here you need to pass the DB connection object as an argument of the finishing method (i.e. where you'd like to fetch from).

ThomasAlban:
Ah yes I hadn't considered how it allows you to force an order in your API. Yeah, makes perfect sense and I retract my previous statement! 
Do you plan on adding something more like a single toggle to reduce API verbosity?

Veetaha:
> I definitely prefer the
> ```rust
> #[builder(start_fn)]
> #[builder(finish_fn)]
> ```

Now I think this notation isn't obvious enough. Maybe this is easier to read:
```rust
#[builder(arg = start_fn)]
#[builder(arg = finish_fn)]
```

Veetaha:
I don't really see a nice syntax for this. The one that sucks less I think is
```
#[builder(source = start_fn)]
#[builder(source = finish_fn)]
```
Which delivers the semantic meaning of "where the value for this member comes from"

ThomasAlban:
Could do 'src' instead of 'source' to make it more concise?

Veetaha:
Yeah, I just can't decide. Now that you put it that way, it seems logical. Glad I can put the blame of naming on someone else 🐱 . Be it `#[builder(start_fn/finish_fn)]`

EdJoPaTo:
I think I prefer `#[builder(on_start_fn)]` to be a bit more clear what is meant from just reading some code someone else write without knowing bon details.

The key=value doesn’t seem to add much helpful context, might suggest more values and is longer so I think having only the keyword is a good choice.

I like thinking about it with multiple people to get different views and thoughts on it!

Veetaha:
> I think I prefer `#[builder(on_start_fn)]` to be a bit more clear what is meant from just reading some code of someone else without knowing bon details.

There is already builder(on(...)) attribute where "on" has a bit different meaning (similar to "if" or "when")

So if adding more clarity, I guess `start_fn_arg` works better.

ThomasAlban:
`start_fn_arg` is also fine, if you think the extra characters are worth it for readability. For me, `start_fn` is just as readable, and I think this feature is going to get used quite a lot so any saving on characters is a benefit!

EdJoPaTo:
Why is saving characters a benefit? Autocompletion exists, looking these up and copying will also be done quite often. The output binary size or performance will not be different in any way. We do not need to save characters. The focus should be on readable code, especially when people read code without being familiar with bon.

Thinking about it again, `#[builder(start_fn)]` might actually be enough as the logic here is "related to builder" → "something with the start function on the builder" So I think its probably fine the way it is then.

Veetaha:
> Why is saving characters a benefit

Just pure aesthetics. 

> Autocompletion exists,

Unfortunately not for proc macros. I experimented with some hacks for autocompletion, and they do work, but they don't work at member level (yet), unfortunately. I'll try to fight with that once there is nothing more in priority for `bon`

Veetaha:
Anyway, all of us are biased because we already know how this feature works. In general, I don't think `_arg` suffix makes that big of a difference. I guess people will either need to look up "what the hell is a `start_fn` and `finish_fn`?". So I'd go with the shorter notation

anelson (approved):

Veetaha:
Done this change was released in a 2.3 version of bon:
- [Reddit post](https://www.reddit.com/r/rust/comments/1fgmbo7/media_nextgen_builder_macro_bon_23_release/)
- [Blog post](https://elastio.github.io/bon/blog/bon-builder-v2-3-release)

RGBCube:
This was merged a while ago, but I actually had to find this to see if positional arguments worked for functions too. It would be great if that example was added to the docs also

Veetaha:
@RGBCube, hi! Thanks for pointing this out. I'm all for improving the docs. Currently, the guide in `bon` [begins](https://bon-rs.com/guide/basics) with this:

> Code examples may use either `fn` or `struct` builder syntax interchangeably. Both follow the same principles unless stated otherwise.

And examples in the [attributes reference](https://bon-rs.com/reference/builder/member/finish_fn) have different "tabs" for different syntax.

image

Basically almost *all* attributes work on all variants of syntax (structs, functions, methods), and I wanted to keep docs shorter and avoid repetitions, so I decided to go with the "tabs" approach in example snippets but only in attributes reference. There could definitely be tabs in the guide section as well, I was just lazy about doing them there because all of them are manually written =)

RGBCube:
Oh, I completely missed that! I was trying to get it done quick and was only skimming the guide itself 😅

I think adding this into the guide would be helpful though, it covers basically everything except this

PR #134, opened 2024-09-14
Make fields name in errors the same as names in struct

andrzejressel:
Also introduces private `mod` space where other elements can be put. This is very useful, because using this we can easily emulate names - `cargo doc` uses local name - so we can have multiple structs that are shown with the same name, but are completelly different.

### DIFF
--- a/bon-macros/src/builder/builder_gen/mod.rs
+++ b/bon-macros/src/builder/builder_gen/mod.rs
@@ -129,7 +129,7 @@ impl Generics {
         }
     }
 
-    fn where_clause_predicates(&self) -> impl Iterator<Item=&syn::WherePredicate> {
+    fn where_clause_predicates(&self) -> impl Iterator<Item = &syn::WherePredicate> {
         self.where_clause
             .as_ref()
             .into_iter()
@@ -147,11 +147,11 @@ impl BuilderGenCtx {
         self.assoc_method_ctx.as_ref()?.receiver.as_ref()
     }
 
-    fn named_members(&self) -> impl Iterator<Item=&NamedMember> {
+    fn named_members(&self) -> impl Iterator<Item = &NamedMember> {
         self.members.iter().filter_map(Member::as_named)
     }
 
-    fn start_fn_args(&self) -> impl Iterator<Item=&StartFnArgMember> {
+    fn start_fn_args(&self) -> impl Iterator<Item = &StartFnArgMember> {
         self.members.iter().filter_map(Member::as_start_fn_arg)
     }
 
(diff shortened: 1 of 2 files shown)


bon/tests/integration/ui/compile_fail/errors.stderr
```diff
@@ -210,14 +210,14 @@ note: required by a bound in `ExampleBuilder::<(__X, __Y, __Z)>::build`
    |            ------- required by a bound in this associated function
    = note: this error originates in the derive macro `Builder` (in Nightly builds, run with -Z macro-backtrace for more info)
 
-error[E0277]: can't finish building yet; the member `ExampleBuilder__renamed` was not set
+error[E0277]: can't finish building yet; the member `renamed` was not set
```
Veetaha, line 213: I'm curious how this works when there are two structs with the fields that are named the same . Won't rustc prepend the module name here to avoid confusion?

bon-macros/src/builder/builder_gen/mod.rs
```diff
@@ -589,12 +589,18 @@ impl BuilderGenCtx {
 
     /// Name of the dummy struct that is generated just to give a name for
     /// the member in the error message when `IntoSet` trait is not implemented.
-    fn members_label(&self, member: &NamedMember) -> syn::Ident {
-        quote::format_ident!(
-            "{}__{}",
-            self.builder_ident.raw_name(),
-            member.setter_method_core_name()
-        )
+    fn members_label(member: &NamedMember) -> &syn::Ident {
```
Veetaha, line 592: I'm not sure why this method was called plural "members" before. I guess it's something I missed during code review. But now I see that it should be just `member_label`, or maybe this method can be removed completely now that it just delegates to a `setter_method_core_name()`.

Same thing about plural naming applies to `absolute_members_label` (it should be `absolute_member_label`).

Veetaha (commented):
LGTM, except for the caveat of two structs using the same field name, I'd like to clarify that. Just want to make sure that `{builder_ident}_inner` module name doesn't unnecessarily leak through the error messages

bon/tests/integration/ui/compile_fail/errors.stderr
andrzejressel: Unfortunately it does. I was testing mostly with `cargo doc` and there are no issues there. I will try to find solution and add it to compile tests.

Veetaha: Yeah, that's unfortunate, and that was one of the reasons why I prepended the builder name to the field here in the first place (although I admittedly, did test my assumption 😳).

Maybe this could be worked around by using a conditional `#[cfg(doc)]`, which I already do as a hack to keep return types of setters shoter in rustdoc output:

https://github.com/elastio/bon/blob/add06f68e5571da4325227757000d2dbb0b48763/bon-macros/src/builder/builder_gen/setter_methods.rs#L190-L194

PR #222, opened 2024-11-28
Support `#[builder(getter(...))]` attribute

### LINKED ISSUE #221: Support getters for already set members
Hi, Bon is my favorite crate and I'm using it as the backbone for a complex pipeline internal framework.

In essence, my pipeline works like this:

```rust
        pipeline
            .pass(the_first_step)
            .pass(another_step)
            .pass(some_other_step)
            .pass(the_last_step)
```

Each step of the pipeline is (simplified) implemented like this:

```rust
pub async fn the_first_step<S>(
    Payload(id): Payload,
    PgConn(pg_conn): PgConn,
    State(builder): State<NewDbRowBuilder<S>>,
) -> Result<NewDbRowBuilder<SetTheFirstStepValue<S>>, crate::Error>
where
    S: new_db_row_builder::State,
    S::TheFirstStepValue: IsUnset,
{
   // some work that connects to api, db, etc, and produces a value

    Ok(builder.the_first_step_value(my_awesome_value_i_fetched))
}
```

The types of the pipeline enforce that the state produced by a prior pass must be fed into the pass under it, so you can't accidentally order anything wrong. I _love_ using Bon to back this pipeline system and allow me to make each pass specify what the conditions of its state should be. 

Where I am struggling is that sometimes I run into a weird, limiting situation.

I have one step in my pipeline that scrapes some information off of a user in my DB, processes it, then sets it in the builder (because this value must go in the row that's being built). Then, in another pass of the pipeline, I must read processed data out and use it to compute another value that also goes on my row. Given the current implementation of Bon (while following the rules and not directly implementing unsafe methods on generated code), I have three solutions:

1. Take this processed value, clone it and set it into the state, then return a tuple of the state and the cloned value. This is bad to me because the passes of my pipeline become much less generic as they now must take a tuple of the built state and any previously obtained values and drill them through the entire way
2. Do what I'm doing currently, and maintain a cache that passes can use to set values and retrieve them later. This works (and is fairly generic), but I lose the ability to _safely_ do the guards like `IsSet` because this cache has to be completely detached from `bon`
3. Use `builder(field)` on these fields I want to cache. This also works, but it has the same problem as before where I lose the ability to do guards on the passes which is what I love so much about bon.

I noticed in 3.1.1 you removed the non-determinism of generated privates. After seeing this, being able to add methods that take in `&self` and return a reference to a value in the builder state would be incredible and make my pipeline work exactly how I want it to. This long winded context finally leads to my question (the one in the title)...

How fundamentally unsafe is this? Taking a look at the expanded macro from `builder` everything I can see indicates that adding read only accessor methods that do not move out of the tuple or clone the data would be completely safe, but I figured it would be well worth opening an issue for this because this will be running in production code where a mistake like this will not be acceptable. 

Thank you!

### A note for the community from the maintainers

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the maintainers with prioritizing it. You may add a comment describing your real use case related to this issue for us to better understand the problem domain.

lazkindness:
Resolves #221 

Implements an experimental (and limited) version of a `getter` attribute to return references to already set values.

Currently supported (for structs, methods, and functions):

- `#[builder(getter)]`
- `#[builder(getter(name = "new_getter_name"))]`
- `#[builder(getter(vis = "pub(crate)"))]`
- `#[builder(getter(doc {}))]`
- ...or any combination of the above 3

In its current form it's not very "smart", so getters for a `String` return `&String` and not `&str`, for example, nor is copy / clone inferred and implemented for types which support them.

See #221 for some more discussion on the future features.


bon-macros/src/builder/builder_gen/member/config/getter.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use super::{docs_utils::parse_docs, SpannedKey};
+
+/// Wrapper around the getter config that allows it to work as both a flag and
+/// a config.
+#[derive(Debug)]
+pub(crate) struct OptionalGetterConfig {
+    span: Span,
+    getter_config: Option<GetterConfig>,
+}
```
Veetaha, line 15: I think you could use `Option<SpannedKey<GetterConfig>>`, where `GetterConfig` is a struct with `name`, `vis` and `docs`. If none of the `name`, `vis` and `doc` were specified (`from_word`), then all those 3 fields would just be `None` (just derive `Default`), so there is no need for a special `Inferred` value or additional `Option` wrapper.

The `SpannedKey` type will contain the span of the `getter` key no matter if it's a simple `getter` or `getter(...)`

bon-macros/src/builder/builder_gen/member/config/getter.rs
```diff
@@ -0,0 +1,99 @@
...
+    Specified(SpecifiedGetterConfig),
+}
+
+impl FromMeta for GetterConfig {
+    fn from_none() -> Option<Self> {
+        None
+    }
+
+    fn from_meta(mi: &syn::Meta) -> darling::Result<Self> {
+        mi.span();
```
Veetaha, line 60: Looks like a leftover

bon-macros/src/builder/builder_gen/setters/mod.rs
```diff
@@ -166,6 +179,29 @@ impl<'a> SettersCtx<'a> {
         Ok([self.setter_method(some_fn), self.setter_method(option_fn)].concat())
     }
 
+    fn getter_method(&self, item: Item) -> TokenStream {
```
Veetaha, line 182: I'd like to avoid mixing code for setters and getters, especially because this file resides in a module called `setters`. I belive it should be in a totally separate `getters` module. Just make sure setters and getters are generated adjacently in the resulting impl block.

bon-macros/src/builder/builder_gen/setters/mod.rs
```diff
@@ -166,6 +179,29 @@ impl<'a> SettersCtx<'a> {
...
+        let state_var = &self.base.state_var;
+        let member_pascal = &self.member.name.pascal;
+        let normalized_type = &self.member.underlying_norm_ty();
+        let state_mod = &self.base.state_mod.ident;
+
+        quote! {
+            #( #docs )*
+            #[allow(clippy::inline_always)]
+            #[inline(always)]
+            #vis fn #name(&self) -> &#normalized_type
```
Veetaha, line 195: We can't guarantee that optional members are set with `Some`, so the current impl is unsound.

For members of the special type `Option<T>` the `underlying_norm_ty()` returns `T`. Instead, for `!member.is_required()` this method needs to return `Option<&T>` and it doesn't need an `unwrap_unchecked` in that case.

bon-macros/src/builder/builder_gen/setters/mod.rs
```diff
@@ -491,18 +533,35 @@ impl SettersItems {
 
         let doc = |docs: &str| iter::once(syn::parse_quote!(#[doc = #docs]));
 
+        let getter_fn = member.config.getter.as_ref().map(|getter_config| Item {
```
Veetaha, line 536: This is also violating single responsibility. Feel free to just copy the relevant parts of this logic into a separate `getters` module.

bon/tests/integration/builder/attr_getter.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[test]
+fn test_struct() {
+    #[derive(Debug, Builder)]
+    #[builder(derive(Debug, Clone))]
+    #[allow(dead_code)]
+
+    struct Sut<T> {
+        #[builder(getter, start_fn)]
+        x1: u32,
```
Veetaha, line 11: Let's add calls to `get_x1`, `x2_with_custom_name`, `get_x3`, `get_generic_option_getter` in the test.

In fact, you'll notice that if you set `maybe_generic_option_getter(None)`, and then call `get_generic_option_getter`, you'll get undefined behavior.

website/src/reference/builder/member/getter.md
```diff
@@ -0,0 +1,87 @@
+# `overwritable` :microscope:
```
Veetaha, line 1:
```suggestion
# `getter` :microscope:
```

Let's also update the table in refererce/builder.md.

You also need to update the sidebar config in `website/.vitepress/config.mts`. Preserve the lexicographical sorting when adding an entry there.

website/src/reference/builder/member/getter.md
```diff
@@ -0,0 +1,87 @@
...
+
+**Applies to:** <Badge type="warning" text="struct fields"/> <Badge type="warning" text="function arguments"/> <Badge type="warning" text="method arguments"/>
+
+Allows getting a reference to an already set member.
+
+::: danger 🔬 **Experimental**
+
+This attribute is available under the cargo feature `experimental-getter`. Breaking changes may occur between **minor** releases but not between patch releases.
```
Veetaha, line 9: Don't forget to add this cargo feature into `Cargo.toml` of `bon` and `bon-macros`, and add validation to the code that reads the config that makes sure that feature is enabled before one could use the `builder(getter)` attribute

website/src/reference/builder/member/getter.md
```diff
@@ -0,0 +1,87 @@
...
+
+The fate of this feature depends on your feedback in the tracking issue [#149](https://github.com/elastio/bon/issues/221). Please, let us know if you have a use case for this attribute!
+
+:::
+
+::: tip
+
+This attribute is also configurable via the top-level [`#[builder(on(...))]`](../top-level/on)
+
+:::
```
Veetaha, line 19:
```suggestion
```

website/src/reference/builder/member/getter.md
```diff
@@ -0,0 +1,87 @@
...
+let builder = Example::builder().x(1);
+
+let x_ref = builder.get_x(); // [!code highlight]
+    
+builder.build();
+```
+
+This enables builders to hold state that can be accessed by the user while still acting as a builder to some final outcome.
+This is especially user if you have a large builder that is getting passed around to many functions,
+and you need to use some previously set value in the builder to set another value.
```
Veetaha, line 39: Let's keep the docs as laconic as possible. `bon`'s documentation is already embarassingly big. I think the use case for getters is already clear, so we can just remove this paragraph. Instead, I'd suggest adding an example of getter for an optional member.

website/src/reference/builder/member/getter.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+## `name`
+
+The default name for getters are chosen according to the following rules:
+
+| Member type | Default                                                   |
+| ----------- | --------------------------------------------------------- |
+| Required    | `get_{member}`                                            |
+| Optional    | `get_{member}`                                            |
```
Veetaha, line 71:
```suggestion
The default name for getter is `get_{member}`
```

lazkindness:
Thanks for the feedback. I'm about done aside from that one comment above, just going to sleep now and finish up the tests in the morning.

bon-macros/src/builder/builder_gen/member/config/getter.rs
Veetaha: Something like this will work. The trick is to manually implement `FromMeta` by using an inner struct that derives `FromMeta` (this trick was used in other places in `bon`, you can find it by "struct Parsed"):

```
#[derive(Default, Debug)]
struct GetterConfig {
    name: Option<SpannedKey<syn::Ident>>,
    vis: Option<SpannedKey<syn::Visibility>>,
    docs: Option<SpannedKey<Vec<syn::Attribute>>>,
}

impl FromMeta for GetterConfig {
    fn from_meta(meta: &syn::Meta) -> Result<Self> {
        if let syn::Meta::Path(_) = meta {
            return Ok(Self::default());
        }

        // Reject empty parens such as `#[builder(getter())]`
        crate::parsing::require_non_empty_paren_meta_list_or_name_value(meta)?;

        // Nested `Parsed` struct used as a helper for parsing the verbose form
        #[derive(FromMeta)]
        struct Parsed {
            name: Option<SpannedKey<syn::Ident>>,
            vis: Option<SpannedKey<syn::Visibility>>,

            #[darling(rename = "doc", default, with = parse_docs, map = Some)]
            docs: Option<SpannedKey<Vec<syn::Attribute>>>,
        }

        let Parsed { name, vis, docs } = Parsed::from_meta(meta)?;

        Ok(Self { name, vis, docs })
    }
}

fn parse_docs(meta: &syn::Meta) -> Result<SpannedKey<Vec<syn::Attribute>>> {
    crate::parsing::parse_docs_without_self_mentions("builder struct's impl block", meta)
}
```

bon-macros/src/builder/builder_gen/getter.rs
```diff
@@ -0,0 +1,106 @@
...
+                // We don't want to avoid using `impl Trait` in the setter. This way
+                // the setter signature is easier to read, and anyway if you want to
+                // specify a type hint for the method that accepts an `impl Into`, then
+                // your design of this setter already went wrong.
+                clippy::impl_trait_in_params,
+                clippy::missing_const_for_fn,
+            )]
+            #[inline(always)]
+            #vis fn #name(&self) -> #return_type
+            where #state_var::#member_pascal: #state_mod::IsSet,
```
lazkindness, line 60: I'm not sure how you feel about this, @Veetaha, but I am deliberately leaving the `IsSet` guard here no matter what even for option type because I think it should be explicitly set to either `Some` or `None` before you can call this.

Veetaha: I agree, I think it makes sense for optional members as well

website/src/reference/builder/member/getter.md
```diff
@@ -68,7 +58,7 @@ The default name for getters are chosen according to the following rules:
 | Member type | Default                                                   |
 | ----------- | --------------------------------------------------------- |
 | Required    | `get_{member}`                                            |
-| Optional    | `get_{member}`                                            |
+| Optional    | `maybe_get_{member}`                                      |
```
Veetaha, line 61: I don't think there is a need to add a `maybe_` prefix for optional members.

website/.vitepress/config.mts
```diff
@@ -323,6 +323,10 @@ export default defineConfig({
                                     text: "setters",
                                     link: "/reference/builder/member/setters",
                                 },
+                                {
+                                    text: "getters",
```
Veetaha, line 327: Let's preserve lexicographical sorting

bon-macros/src/builder/builder_gen/getter.rs
```diff
@@ -0,0 +1,106 @@
...
+        quote! {
+            #( #docs )*
+            #[allow(
+                // This is intentional. We want the builder syntax to compile away
+                clippy::inline_always,
+                // We don't want to avoid using `impl Trait` in the setter. This way
+                // the setter signature is easier to read, and anyway if you want to
+                // specify a type hint for the method that accepts an `impl Into`, then
+                // your design of this setter already went wrong.
+                clippy::impl_trait_in_params,
```
Veetaha, line 55: I don't think ignoring this lint is needed for getters, also the comment here is stale.

bon-macros/src/builder/builder_gen/getter.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+impl GetterItem {
+    fn new(ctx: &GetterCtx<'_>) -> Option<Self> {
+        let GetterCtx { member, base } = ctx;
+
+        let Some(spanned_keyed_config) = member.config.getter.as_ref() else {
+            return None;
+        };
```
Veetaha, line 74:
```suggestion
        let spanned_keyed_config = member.config.getter.as_ref()?;
```
Also, `let-else` isn't available in `bon` at the moment do to `1.59` MSRV (let-else appeared in 1.65). Otherwise a `?` works here anyway.

Veetaha (commented):
LGTM, although I did just a quick review this time. Once new comments are fixed I'll merge this. I'll re-review the changes tomorrow, improve any other minor stuff, maybe update some of the guides, and do the minor release.

website/src/reference/builder/member/getter.md
```diff
@@ -0,0 +1,77 @@
+# `getter` :microscope:
```
Veetaha, line 1: One more thing. There is a table of all attributes in `website/src/reference/builder.md`, it should be updated

lazkindness:
> LGTM, although I did just a quick review this time. Once new comments are fixed I'll merge this. I'll re-review the changes tomorrow, improve any other minor stuff, maybe update some of the guides, and do the minor release.

Cool, I do also have a need for the `copy` and `clone` attributes so some time this weekend or next week I can probably take a stab at those if you don't get to it already

lazkindness:
Pipeline should be good now too, I added a tiny bit to the init script and contributing docs

Veetaha:
I'm pushing my corrections and merging

bon-macros/src/builder/builder_gen/member/config/mod.rs
```diff
@@ -212,15 +224,42 @@ impl MemberConfig {
...
+        }
+
+        if let Some(getter) = &self.getter {
+            if !cfg!(feature = "experimental-getter") {
+                bail!(
+                    &getter.key.span(),
+                    "`getter` attribute is experimental and requires \
+                    \"experimental-getter\" cargo feature to be enabled; \
+                    we would be glad to make this attribute stable if you find it useful; \
+                    please leave a 👍 reaction under the issue https://github.com/elastio/bon/issues/221 \
```
beyera, line 246:
```suggestion
                    please leave a 👍 reaction under the issue https://github.com/elastio/bon/issues/225 \
```

bon/Cargo.toml
```diff
@@ -82,3 +82,17 @@ implied-bounds = ["bon-macros/implied-bounds"]
...
+# 🔬 Experimental! There may be breaking changes to this feature between *minor* releases,
+# however, compatibility within patch releases is guaranteed though.
+#
+# This feature enables the #[builder(getter)] attribute that can be used to
+# allow getting references to already set fields in the builder.
+#
+# See more info at https://bon-rs.com/reference/builder/member/getter.
+#
+# We are considering stabilizing this attribute if you have a use for it. Please leave
+# a 👍 reaction under the issue https://github.com/elastio/bon/issues/221 if you need
```
beyera, line 95: Same here? 

```suggestion
# a 👍 reaction under the issue https://github.com/elastio/bon/issues/225 if you need
```

bon-macros/src/builder/builder_gen/member/config/mod.rs
Veetaha: Yeah, I update this code locally, although I can't push my changes to [Kindness-Works:laz/221-getters](https://github.com/Kindness-Works/bon/tree/laz/221-getters) even though there is this sentence in the PR metadata:
image

Maybe you have an org-wise rejection for writes from external contributors?

Anyway, I'll force-merge this PR and submit my amendments in a separate PR

Veetaha:
Openned a followup PR https://github.com/elastio/bon/pull/226

Veetaha:
This  feature has been released in [bon 3.2.0](https://github.com/elastio/bon/releases/tag/v3.2.0)

Veetaha:
@lazkindness I opened an (unfinished yet) draft PR for `clone`, `copy`, `deref` params support: https://github.com/elastio/bon/pull/229.

Veetaha:
Heads up, did a `3.3.0` release with `copy`, `clone`, `deref` support: https://github.com/elastio/bon/releases/tag/v3.3.0.

cc @lazkindness

Issue #230, opened 2024-12-06
Add the ability to change `maybe_` for optional fields 


MuhannadAlrusayni:
Hi there,  

First off, thank you for creating and maintaining Bon—it’s been a pleasure working with it!  

As I migrate a codebase to Bon, I’ve run into a challenge with the naming conventions for setters handling `Option` fields. My current codebase uses `set_` as the prefix for these setters. While Bon's built-in support for `Option` setters is a fantastic feature that removes the need for manual definitions, it uses the `maybe_` prefix by default.  

This mismatch means I still have to manually rename every setter to match my convention, which undermines some of the convenience.  

I’d like to suggest two potential improvements to make this process smoother for me and other users:  
1. **Set `set_` as the default prefix for `Option` setters**
2. **Introduce configurability for setter prefixes at the type level:**  
   Allowing users to define their preferred prefix (e.g., `set_`, `maybe_`, or custom ones) would offer greater flexibility and make Bon more adaptable to different codebases and conventions.

While either of the suggested changes would address my use case, I believe adding the ability to configure this at the type level would provide greater flexibility and cover a wider range of use cases.

**Code snippet:**
```rust
#[derive(Builder)]
#[builder(on(optional, prefix = "set_"))] // Example syntax for explanation, not necessarily the suggested approach
struct Foo {
    value: Option<String>,
    label: Option<String>,
}
```

### A note for the community from the maintainers

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the maintainers with prioritizing it. You may add a comment describing your real use case related to this issue for us to better understand the problem domain.

Veetaha:
Hi, thank you for creating the issue! Overriding the `maybe_` prefix is indeed something that I've been thinking of having in `bon`.

I suppose as a workaround you use `#[builder(setters(option_fn(name = set_{member_name})))]`? That's indeed rather inconvenient. Having a prefix supported via `#[builder(on(...))]` would be preferred.

I have a big and complex issue about non-flag attributes support in `#[builder(on(...))]` (https://github.com/elastio/bon/issues/152) with a comprehensive design for precedence and conflicts. However, I don't think this feature would have to be blocked by the full implementation of #152, because I believe overriding the `maybe_` prefix may be a frequent use case. We can have a limited, but still useful implementation that accepts only this syntax:

```rust
#[builder(on(_, setters(option_fn(prefix = set_))))]
```

[...]

PR #233, opened 2024-12-12
Support name `prefix` configs for setters

Veetaha:


Veetaha:
Closing this PR since I've found an alternative, better way to do top-level config of the method names using a template string with `{}` placeholder (https://github.com/elastio/bon/issues/230#issuecomment-3893510813). Thanks @dflemstr for this idea, which based on the design of #364

PR #282, opened 2025-04-15
Add support for propagating `#[track_caller]` and future special attributes

lilyyy411:
As discussed on Discord, propagates `#[track_caller]` on `#[builder]` functions, along with some tiny internal refactors to allow propagating other special attributes in the future.


Veetaha (approved):
Looks good, thank you! I'll make a new patch after this is merged

PR #284, opened 2025-04-16
Propagate `#[target_feature]` attributes on `#[builder]`-generated finishing functions

lilyyy411:
As discussed on Discord, propagates `#[target_feature]` attributes on `#[builder]` functions, and replaces `#[inline(always)]` on the outer builder function with `#[inline]` when there is a `#[target_feature]` attribute detected.

For those who are unaware about why this is needed, [Rust 1.86.0](<https://blog.rust-lang.org/2025/04/03/Rust-1.86.0/>) added the ability to apply `#[target_feature]` to non-`unsafe` functions. The function becomes safe to call in other functions marked with the same target feature, but are `unsafe` to call otherwise, breaking `bon` builders.


bon/tests/integration/builder/target_feature.rs
```diff
@@ -0,0 +1,42 @@
+#[cfg(target_arch = "x86_64")]
+#[rustversion::since(1.86.0)]
+mod msrv_1_86 {
+    #![allow(dead_code)]
+    use crate::prelude::*;
+
+    #[test]
+    fn target_feature_fn() {
```
Veetaha, line 8: `[nit]`

By convention, the test functions are called `test_struct`, `test_function` and `test_method` for different variants of syntax.

bon/tests/integration/builder/target_feature.rs
```diff
@@ -0,0 +1,42 @@
...
+                let Brick(_) = brick;
+                Self
+            }
+        }
+
+        #[target_feature(enable = "avx2")]
+        #[allow(unsafe_code)]
+        unsafe fn briiick() {
+            Senti::builder().brick(Brick([0; 32])).yatta_but_wide();
+        }
```
Veetaha, line 40: Let's put this under the impl block too, otherwise it's testing the same thing that the free-function test higher already does

Veetaha (approved):
Thank you for the contribution! Since this isn't critical, I'll release this in a new patch on the next maintenance pass next Friday with any other changes that may land by that time

PR #313, opened 2025-07-03
Adding support for `build_from` and `build_from_clone` builder methods.

### LINKED ISSUE #310: Feature Request: Add .from(&T) method to Builder for cloning from an existing instance.
I'd like to propose support for a `.from(&T)` method on `bon` builders that prepopulates the builder fields by cloning from an existing instance. This would streamline a common use case such as updating immutable database entities in web applications.

**Use Case**

This is especially useful in web servers: a struct is loaded from the database, a few fields are updated, and the new value is persisted. Currently, this requires repetitive field copying:

```rust
let updated = User::builder()
    .id(user.id.clone())
    .email(user.email.clone())
    .username("new_username".into())
    .build();
```

With `.from(&user)`:

```rust
let updated = User::builder()
    .from(&user)
    .username("new_username".into())
    .build();
```

**Proposal**

- Add `.from(&T)` to the builder, opt-in via `#[builder(from_instance)]`.
  -  `#[builder(from_instance)]` can ensure that `Clone` is derived as well.
- Clone each field from the input value.
- Skip fields marked `#[builder(skip)]`.
- Maintain `bon`'s type-state guarantees.

**Benefits**

- Reduces boilerplate for partial updates.
- Aligns with real-world workflows (e.g., editing DB records).
- Keeps compile-time safety and existing behavior intact.

Let me know if this is something you'd consider. I'd be happy to assist or prototype it.

Thank you for making `bon`, it has saved me so many hours of writing Rust boilerplate code!

BernardIgiri:
# Add `build_from` and `build_from_clone` builder methods

Closes #310

This PR adds support for two new optional top-level builder attributes:

* `#[builder(build_from)]` – Adds a `.build_from(&T)` method
* `#[builder(build_from_clone)]` – Adds a `.build_from_clone(T)` method

These methods allow partially configured builders to fill in missing fields from an existing instance of the target type (`T`) before finalizing the build. This is useful when:

* You want to override only a few fields but reuse most of the existing data.
* You're working with types that aren't `Default`, making field reuse non-trivial.

## Example

```rust
#[derive(Builder, Clone)]
#[builder(build_from, build_from_clone)]
struct User {
    name: String,
    age: u8,
}

let jon = User::builder().name("Jon".into()).age(25).build();
let alice = User::builder().name("Alice".into()).build_from(&jon);
assert_eq!(alice.name, "Alice");
assert_eq!(alice.age, 25);
```

## Implementation notes

* The new code lives in `builder_gen/build_from.rs` and is only emitted if the attributes are enabled.
* Each field is handled according to its member kind:

  * `Named` fields fallback to `from.field.clone()` if not explicitly set.
  * `FinishFn` fields always defer to `from.field.clone()`.
  * `Skip` fields use `Default::default()`.
* Integration tests verify both methods.

## Next steps

* Reference docs have been stubbed in `src/__/ide.rs`; we may want to flesh out the corresponding links on the site.

@Veetaha This implements the optional build_from and build_from_clone methods discussed in #310 — feedback welcome!


bon-macros/src/builder/builder_gen/top_level_config/mod.rs
```diff
@@ -75,6 +75,12 @@ pub(crate) struct TopLevelConfig {
     /// Specifies the derives to apply to the builder.
     #[darling(default, with = crate::parsing::parse_non_empty_paren_meta_list)]
     pub(crate) derive: DerivesConfig,
+
+    #[darling(default)]
+    pub(crate) build_from: bool,
```
Veetaha, line 80: These should use `Option<Option<ItemSigConfig>>` to make their names, visibility and docs configurable, plus inherit the visibility of the builder struct similar to how the regular `finish_fn` does that, otherwise the docs would be incorrectly claiming that these attributes support `vis`, `name`, `doc` attributes.

I think this may require writing a custom `with` parser function that accepts a `syn::Meta::Path` (in which case `None` (inner None), and otherwise delegates to `ItemSigConfig`.

bon-macros/src/builder/builder_gen/build_from.rs
```diff
@@ -0,0 +1,133 @@
...
+            match member {
+                Member::Field(_) | Member::StartFn(_) => quote! {
+                    let #ident: #ty = self.#ident;
+                },
+                Member::Named(member) => {
+                    let index = &member.index;
+                    quote! {
+                        let #ident: #ty = match self.__unsafe_private_named.#index {
+                            Some(value) => value,
+                            None => from.#ident.clone(),
```
Veetaha, line 27: The idea is that `build_from` accepts `T` by value and does not call `clone()` at all (that's why this method's name is a bit shorter since it is less expensive and should be the preferred thing), but `build_from_clone` accepts `&T` and calls `.clone()` if needed (that's why it has "clone" in its name).

bon-macros/src/builder/builder_gen/build_from.rs
```diff
@@ -0,0 +1,133 @@
...
+    })?;
+
+    let ident = path
+        .segments
+        .last()
+        .ok_or_else(|| err!(&span, "expected a named type, but found an empty path"))?
+        .ident
+        .clone();
+
+    Ok(quote_spanned! { span => #ident })
```
Veetaha, line 132: This is a known footgun of `quote_spanned!`. It doesn't overwrite the span of interpolated tokens such as `#ident` - only the bare tokens created as part of that quasi-quotation syntax. But, otherwise, I don't see a need to override the span here. The span is preserved from the input by default, and that's desired

bon-macros/src/builder/builder_gen/build_from.rs
```diff
@@ -0,0 +1,133 @@
...
+    let ctor_path = extract_ctor_ident_path(target_ty, target_ty.span())?;
+
+    Ok(quote! {
+        #[inline(always)]
+        #[doc = #doc]
+        pub fn #method_name(self, #arg_pat: #arg_type) -> #target_ty {
+            #( #field_vars )*
+            #ctor_path {
+                #( #ctor_args, )*
+            }
```
Veetaha, line 110: Unfortunatelly, this approach won't work in case if the builder is generated with the `#[builder]` macro on top of a function. We have the `body.generate()` in `finish_fn.rs` here that generates the proper body: https://github.com/elastio/bon/blob/af6e9251ec65217e974ec3dcc447f9c1e456cd72/bon-macros/src/builder/builder_gen/finish_fn.rs#L139

It's only requirement is that members are assigned to variables of the same name in scope. Plus we need to inherit the potential `async` / `unsafe` / `const` modifiers and maybe handle the case of the `Result` from the original function

Veetaha (commented):
There are a bunch of (quite a lot actually) edge cases that this needs to handle. Not saying we should handle them all right now, we can start with a good enough MVP and put it under the `experimental-build-from` flag for now and iterate further.

I'm still not feeling in love with the attribute names and syntax. I also considered nesting them under finish_fn like `#[builder(finish_fn(build_from))]` to reduce noise in the site docs, but this syntax doesn't look intuitive (seems like we are overriding the `finish_fn` instead of adding one more method. Btw. we should also update them (they are in the `website` directory in this repo). But IDK, I may change the naming if some great idea comes to mind a bit later

Could also be done via `#[builder(derive(BuildFrom))]` like it's a "derive" for a builder that adds one more method. Still not sure

polarathene:
@BernardIgiri any reason for closing the PR? You opened it, received feedback almost a year ago, but chose not to continue? Did it seem like too much effort?

Or was this unintentional? It looks like you pushed the `master` branch to this PR branch, as the PR diff is now empty (_which would close the PR automatically as a result_), I'm assuming that was a mistake and you intended to rebase your PR to the latest `master` branch?

BernardIgiri:
> @BernardIgiri any reason for closing the PR? You opened it, received feedback almost a year ago, but chose not to continue? Did it seem like too much effort?
> 
> Or was this unintentional? It looks like you pushed the `master` branch to this PR branch, as the PR diff is now empty (_which would close the PR automatically as a result_), I'm assuming that was a mistake and you intended to rebase your PR to the latest `master` branch?

Sorry, I was trying to rebase so I can update this PR. I made good progress last night but might need some help. I will reopen this PR this evening on EST time after work today.

BernardIgiri:
Okay. I created a new PR since reopening this seemed harder to do. I will continue this work there.

Issue #314, opened 2025-07-05
Feature Request: IntoFuture


KaiserKarel:
Reqwest and other networking libraries often have their request structs implement [`IntoFuture`](https://doc.rust-lang.org/std/future/trait.IntoFuture.html), making the following pattern possible:

```
let result = Request::new().await

let result = Request::new().with_something(x).await
```

It'd be nice if functions which return a Future, and are decorated with bon, would result in builders that implement IntoFuture.

Veetaha:
Closed by #322. Released in [`bon v3.7.0`](https://github.com/elastio/bon/releases/tag/v3.7.0).

PR #322, opened 2025-08-09
Add support for `derive(IntoFuture(Box))`

jakubadamw:
Addresses #314.

This allows for a derivation of the `IntoFuture` trait which enables a pattern by which a builder holding all required properties and ready to be finalised, can be `await`ed on to produce the final result asynchronously.


jakubadamw:
> Thanks for the contribution. I pushed some improvements to this PR as well as support for non-`'static` lifetimes. CI seems to be failing on nightly/beta toolchains for unrelated reasons. I'll fix those separately and push a new minor release shortly.

Thank you for the follow-up work and merging the PR. 7c9f53a96b226828a8a0d2191fef868e3cd005cf looks like it addressed a significant omission, in particular.

PR #331, opened 2025-08-19
Add Copyright header to the MIT License

Veetaha:
The MIT license has a standard header for the copyright notice. Apache doesn't seem to have one. Anyway, I think this should be enough

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,3 +1,7 @@
+The MIT License (MIT)
+
+Copyright (c) 2024 Elastio Software Inc
+
 Permission is hereby granted, free of charge, to any
 person obtaining a copy of this software and associated
 documentation files (the "Software"), to deal in the


anelson (approved):

WeepingClown13:
@Veetaha this looks good, thanks a lot :) btw why not 2025? copyright years are not necessary in a copyright, but since you have it and are just starting, might as well go with the latest year.

Veetaha:
@WeepingClown13 

2024 is the year this work started and was made public, which is what the year should stand for according to [Wiki](https://en.wikipedia.org/wiki/Copyright_notice)🤔

Use of the notice informs the public that a work is protected by copyright, identifies the copyright owner, and shows the year of first publication.

WeepingClown13:
@Veetaha as I said, copyright years are not mandatory. If you have 2024, then 2024-current is inferred. If you have 2025, then we take it that you have the copyright in the latest year. Even if you not add anything, it still doesn't matter as you own the copyright anyway since the first publishing of the work. So it's nothing to be hung up on. It's your call, doesn't really matter :)

Veetaha:
Okay, thanks for looking at this. The year doesn't seem very important, so I'll follow Wiki's stance on this here. Merging, and preparing a patch

WeepingClown13:
@Veetaha thank you!

an example for some further reading: https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/

Veetaha:
Huh, thanks for sharing that. Curl's view on this is really useful. It's also the reason I intentionally avoided a range of years here =)

PR #336, opened 2025-09-02
Set lowerbound of `rustversion` of to v1.0.19

Fokko:
I got the following error in Iceberg-Rust:

```
   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bon-macros-3.7.0/src/builder/builder_gen/getters.rs:227:46
    |
227 |                 let module = if rustversion::cfg!(since(1.64.0)) {
    |                                              ^^^ could not find `cfg` in `rustversion`

```

Which depends on Avro-rs, and Avro-rs depends on Bon.

It looks like `rustversion::cfg!` has been added in 1.0.19:

- https://crates.io/crates/rustversion/1.0.19
- https://crates.io/crates/rustversion/1.0.18

### DIFF
--- a/bon-macros/Cargo.toml
+++ b/bon-macros/Cargo.toml
@@ -58,7 +58,7 @@ quote = "1"
 syn = { version = "2.0.56", features = ["full", "visit-mut", "visit"] }
 
 prettyplease = "0.2"
-rustversion  = "1.0"
+rustversion  = "1.0.18"
 
 [features]
 default = []


Veetaha (approved):
Thank you for the fix! I double checked with `cargo +nightly update -Z minimal-versions` and this seems to be the only offender. Will release this as 3.7.2.

**UPD:** already [released](https://bon-rs.com/changelog#_3-7-2-2025-09-02).

Fokko:
Wow, thanks for the quick follow-up, much appreciated 👍

PR #341, opened 2025-10-02
Preserve `fn` and `{}` for `start_fn` so that rustdoc generates correct links

### LINKED ISSUE #57: Source references in docs generated by rustdoc lead to the macro callsite instead of the original function
If you click on any `Source` link for the methods in [this library](https://docs.rs/remoteit-api/latest/remoteit_api/struct.R3Client.html#method.get_files_async), the line of code that is opened is the line of `#[bon]` macro call site, not the original function itself.

Instead, the docs should forward to the function that was used to generate the builder.

Eisverygoodletter:
## Preserve `fn` and `{}` during `start_fn` construction
Closes #57.

This PR changes the generation of `start_fn` such that it reuses the `fn` keyword and `{` `}` tokens from the original definition.

This fix works because rustdoc only considers the spans of the first and last tokens when determining the span of a function. In our case, the start token is `pub`, `const` or `fn`. The ending token is `}`. The original code already preserves `pub` and `const` so I have added code for preserving `fn` and `{}`.

The original implementation relies on `quote!` for generating `{`, `}` and `fn` which default to `Span::call_site()` which is why rustdoc linked to the macro callsite instead of the original function.

## Seeing the changes
Some methods in the bon sandbox such as https://github.com/elastio/bon/blob/c0e4356d174faf108101fbb9141effbd06980ced/bon-sandbox/src/docs_comparison/methods.rs#L10-L26 should now generate a correct source link that links to the original function. Can be tested using `cargo doc --package bon-sandbox --open`.

## Possible future improvements
Since rustdoc only cares about the start and end tokens, it is theoretically possible to modify builder method generation such that it also reuses the spans of the original struct member/parameter definitions (not implemented in this PR). This would be more complicated because you would have to assign spans from one type of token to another type of token which I haven't tested yet. I might make an issue for that later after digging through more of the codebase and thinking a bit more.

### DIFF
--- a/bon-macros/src/builder/builder_gen/models.rs
+++ b/bon-macros/src/builder/builder_gen/models.rs
@@ -86,12 +86,8 @@ pub(super) struct StartFn {
     /// Overrides the default generics
     pub(super) generics: Option<Generics>,
 
-    /// Preserve the original fn token span.
-    /// This determines the generated fn's span start
-    pub(super) fn_token: syn::token::Fn,
-    /// Preserve the original brace token.
-    /// This determines the generated fn's span end
-    pub(super) brace_tokens: syn::token::Brace,
+    /// Preserve a span the documentation for `start_fn` should link to.
+    pub(super) span: Span,
 }
 
 pub(super) struct StartFnParams {
@@ -104,12 +100,9 @@ pub(super) struct StartFnParams {
 
     /// Overrides the default generics
     pub(super) generics: Option<Generics>,
-    /// Preserve original fn token for its span if possible. Defaults to
-    /// macro callsite if not supplied.
-    pub(super) orig_fn_token: Option<syn::token::Fn>,
-    /// Preserve original brace tokens `{}` for their spans if possible.
+    /// Preserve a span the documentation for `start_fn` should link to.
     /// Defaults to macro callsite if not supplied.
-    pub(super) orig_brace_tokens: Option<syn::token::Brace>,
+    pub(super) span: Option<Span>,
 }
 
 pub(super) struct BuilderType {
@@ -322,8 +315,7 @@ impl BuilderGenCtx {
             vis: start_fn.vis.unwrap_or_else(|| builder_type.vis.clone()),
             docs: start_fn.docs,
             generics: start_fn.generics,
-            fn_token: start_fn.orig_fn_token.unwrap_or_default(),
-            brace_tokens: start_fn.orig_brace_tokens.unwrap_or_default(),
+            span: start_fn.span.unwrap_or_else(Span::call_site),
         };
 
         let finish_fn = FinishFn {
--- a/bon-macros/src/builder/builder_gen/models.rs
+++ b/bon-macros/src/builder/builder_gen/models.rs
@@ -88,10 +88,10 @@ pub(super) struct StartFn {
 
     /// Preserve the original fn token span.
     /// This determines the generated fn's span start
-    pub(super) orig_fn_token: syn::token::Fn,
+    pub(super) fn_token: syn::token::Fn,
     /// Preserve the original brace token.
     /// This determines the generated fn's span end
-    pub(super) orig_brace_tokens: syn::token::Brace,
+    pub(super) brace_tokens: syn::token::Brace,
 }
 
 pub(super) struct StartFnParams {
@@ -322,8 +322,8 @@ impl BuilderGenCtx {
             vis: start_fn.vis.unwrap_or_else(|| builder_type.vis.clone()),
             docs: start_fn.docs,
             generics: start_fn.generics,
-            orig_fn_token: start_fn.orig_fn_token.unwrap_or_default(),
-            orig_brace_tokens: start_fn.orig_brace_tokens.unwrap_or_default(),
+            fn_token: start_fn.orig_fn_token.unwrap_or_default(),
+            brace_tokens: start_fn.orig_brace_tokens.unwrap_or_default(),
         };
 
         let finish_fn = FinishFn {
--- a/bon-macros/src/builder/builder_gen/models.rs
+++ b/bon-macros/src/builder/builder_gen/models.rs
@@ -85,6 +85,13 @@ pub(super) struct StartFn {
 
     /// Overrides the default generics
     pub(super) generics: Option<Generics>,
+
+    /// Preserve the original fn token span.
+    /// This determines the generated fn's span start
+    pub(super) orig_fn_token: syn::token::Fn,
+    /// Preserve the original brace token.
+    /// This determines the generated fn's span end
+    pub(super) orig_brace_tokens: syn::token::Brace,
 }
 
 pub(super) struct StartFnParams {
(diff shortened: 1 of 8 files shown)


bon-macros/src/builder/builder_gen/models.rs
```diff
@@ -85,6 +85,13 @@ pub(super) struct StartFn {
 
     /// Overrides the default generics
     pub(super) generics: Option<Generics>,
+
+    /// Preserve the original fn token span.
+    /// This determines the generated fn's span start
+    pub(super) orig_fn_token: syn::token::Fn,
+    /// Preserve the original brace token.
+    /// This determines the generated fn's span end
+    pub(super) orig_brace_tokens: syn::token::Brace,
```
Veetaha, line 94: `orig` in the field name couples the `StartFn` model with the `fn`-based builder generation logic logically. Since this struct is reused between `struct`-based and `fn`-based builders lets keep it agnostic to the origin of the builder. I think it's enough to just strip the `orig_` prefix from field names here. Just `fn_token` and `brace_tokens` is fine as it severs as a "request" to use these tokens for the generated starting function.


```suggestion
    pub(super) fn_token: syn::token::Fn,
    /// Preserve the original brace token.
    /// This determines the generated fn's span end
    pub(super) brace_tokens: syn::token::Brace,
```

bon-macros/src/builder/builder_gen/start_fn.rs
```diff
@@ -109,24 +132,12 @@ impl super::BuilderGenCtx {
                 // const operations.
                 clippy::missing_const_for_fn,
             )]
-            #vis #const_ fn #start_fn_ident< #(#generics_decl),* >(
+            #vis #const_ #fn_kw #start_fn_ident< #(#generics_decl),* >(
```
Veetaha, line 135: Note that an interesting positive byproduct of this change is that rustc started showing warnings for unused builder methods in `#[bon]` impl blocks. I didn't realise it wasn't actually showing these warnings. There are some tests in this repo that define but do not use some builder methods. This is done either intentionally (e.g. the test only validates the code compiles), or unintentionally (some asserts were missing in tests). These need to be fixed, otherwise CI will fail with `dead_code` errors. Feel free to add `allow`s for any new warnings, or add assertions that use the unused methods in our test

Veetaha (commented):
Didn't expect this to be that simple. Thank you for researching and fixing this! Awaiting for tests update before merge

bon-macros/src/builder/builder_gen/models.rs
Eisverygoodletter: I have removed the `orig_` prefix.

Eisverygoodletter: Just realised I missed a few spots, will be fixing soon.

Eisverygoodletter: Changed to an even simpler implementation, the code here is no longer needed

bon-macros/src/builder/builder_gen/start_fn.rs
```diff
@@ -96,6 +96,29 @@ impl super::BuilderGenCtx {
...
+                #( #custom_fields_idents, )*
+                #receiver_field_init
+                #( #start_fn_args_fields_idents, )*
+                __unsafe_private_named: #named_members_field_init,
+            }
+        };
+        let body_block = syn::Block {
+            brace_token: self.start_fn.orig_brace_tokens,
+            stmts: body_contents,
+        };
```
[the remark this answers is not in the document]
Eisverygoodletter: Result with `syn::parse_quote_spanned!` was good so I have done a rewrite of my implementation. Resolving this thread because I changed how I preserve spans (thread outdated and suggestion for using spans kinda applied)

bon-macros/src/builder/builder_gen/start_fn.rs
Veetaha: I think it's okay to ignore this lint. It's not always possible to elide the lifetime like that what clippy suggests.

For example, here is a case:

```rust
#[bon::builder]
fn example<'a>(
    #[builder(with = |value: &'a str| value.trim())]
    x: &'a str
) {
    let _ = x;
}
```

Here, clippy suggests eliding the lifetime, but this won't work because the lifetime is used in the closure's signature. The problem is that this code generates several different items, and clippy doesn't know that `builder` macro uses the lifetime name to establish a correlation between lifetimes in the function with the lifetime names in the generated setters (that use the signature from `with` closure).
 
 
For this reason I've been thinking of unconditionally `allow`-ing this specific clippy lint about lifetime elision to prevent false positives such as this example. I haven't gotten to do it because I thought it'd hide some true positives of this lint (where lifetime elision would actually work) and `bon` would be guilty of hiding that from the developer. However, now I think it's not big of a deal, so feel free to add that `alllow` to the generated code

Eisverygoodletter: Made a commit so that `allow(clippy::elidable_lifetime_names)` is now emitted alongside other annotations. I've run `cargo clippy --all-targets --locked --features experimental-overwritable` locally and it doesn't emit any warnings, so hopefully the CI should pass now.

Eisverygoodletter:
Just checked the CI results, seems like I blocked out `clippy::elidable_lifetime_names` (pedantic-allow lint) but `clippy::needless_lifetimes` (complexity-warn lint) is triggering. Not sure why it's only failing on MSRV tests, maybe they changed the lints sometime between versions? I'll try test it locally.

bon-macros/src/builder/builder_gen/start_fn.rs
```diff
@@ -96,8 +96,31 @@ impl super::BuilderGenCtx {
...
+            quote! {
+                #[allow(clippy::needless_lifetimes)]
+            }
+        }
+        #[rustversion::since(1.87)]
+        fn get_needless_lifetime_lint_annotation() -> TokenStream {
+            quote! {
+                #[allow(clippy::elidable_lifetime_names)]
+            }
+        }
```
Veetaha, line 117: Could be simplified slightly to avoid some repetition like this:

```rust
        let needless_lifetime_lint = if rustversion::cfg!(before(1.87)) {
            format_ident!("needless_lifetimes")
        } else {
            format_ident("elidable_lifetime_names")
        };
        
        // inside final syn::parse_quote_spanned
        #[allow(clippy::#needless_lifetime_lint)]
```

Veetaha (approved):

Eisverygoodletter:
I don't actually think that the lints were triggered by `bon` generated code. My original code (before bon macro application) was already triggering the false positives so it isn't `bon`'s problem.

I tested with a few more examples and it seems like the issues are caused by internal clippy bugs ([link to github issues page with relevant tags](https://github.com/rust-lang/rust-clippy/issues?q=is%3Aissue%20needless_lifetimes%20label%3AC-bug)) making the lint trigger on some false positives. The lint issues I mentioned were my own (local) test cases which was why they triggered the niche false positives. 

We don't need to do anything for the lint problems I mentioned, the CI is right.

Veetaha:
Alright, thank you for the contribution! I'll create a new patch with this after merge

PR #382, opened 2026-04-16
Replace `cargo-asm` with `cargo-show-asm`

pacak:
Use a working alternative. Chances are `cargo-asm` is not going to work with more up to date rust compiler.

Invocation for the script is the same, so it should just work. `--no-color` is not needed for redirects.


Veetaha:
Thanks for the heads up. Not sure why I didn't find `cargo-show-asm` earlier. And a big thank you for this tool!

Veetaha (approved):

PR #385, opened 2026-06-10
Fix rust-analyzer "Go to Definition" inside `#[builder]` functions

skreborn:
Rust-analyzer's "Go to Definition" function is currently broken when used inside a function annotated with `#[builder]` because of [the way it prioritizes spans](https://github.com/rust-lang/rust-analyzer/issues/18438).

Prior to this change, using "Go to Definition" on any token within the function body led the analyzer wildly astray. After the fix, it correctly jumps to the item definition.

As far as I can tell, this change doesn't introduce any regressions or changes in generated documentation.

### DIFF
--- a/bon-macros/src/builder/builder_gen/start_fn.rs
+++ b/bon-macros/src/builder/builder_gen/start_fn.rs
@@ -109,10 +109,7 @@ impl super::BuilderGenCtx {
             format_ident!("elidable_lifetime_names")
         };
 
-        // Construct using a span which links to our original implementation.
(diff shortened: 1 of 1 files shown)


Veetaha (approved):
I've checked this change, and it seems to work fine. Source links in rustdoc still seem to work correctly. I'll make a patch release asap. Thanks for the PR!

Veetaha:
This change was released in [v3.9.2](https://github.com/elastio/bon/releases/tag/v3.9.2)

Issue #152, opened 2024-10-20
`#[builder(on(...))]` extensions


Veetaha:
Currently, there is a `#[builder(on(pattern, attrs))]` attribute that supports `into` configuration. With that one it's possible to enable `#[builder(into)]` for all members like this:

```rust
use std::path::{Path, PathBuf};

#[derive(bon::Builder)]
#[builder(on(_, into))]
struct Example {
    // auto #[builder(into)]
    x1: String,
    // auto #[builder(into)]
    x2: PathBuf,
    // auto #[builder(into)]
    x3: Box<str>,
    // auto #[builder(into)]
    x4: Box<Path>,
}
```

The `_` is a pattern that matches all members. However, it's possible to match only specific types like this:

```rust
#[derive(bon::Builder)]
#[builder(on(String, into))]
struct Example {
    // auto #[builder(into)]
    x1: String,
    x2: PathBuf,
    x3: Box<str>,
    x4: Box<Path>,
}
```

This way only `x1` will get a `#[builder(into)]` because it's type `String` matches the pattern. To match any member of `Box` type the following pattern could be used:

```rust
#[derive(bon::Builder)]
#[builder(on(Box<_>, into))]
struct Example {
    x1: String,
    x2: PathBuf,
    // auto #[builder(into)]
    x3: Box<str>,
    // auto #[builder(into)]
    x4: Box<Path>,
}
```

It's also possible to specify multiple `on(...)` clauses:

```rust
#[derive(bon::Builder)]
#[builder(
    on(Box<_>, into),
    on(String, into),
)]
struct Example {
    // auto #[builder(into)]
    x1: String,
    x2: PathBuf,
    // auto #[builder(into)]
    x3: Box<str>,
    // auto #[builder(into)]
    x4: Box<Path>,
}
```

This syntax of `#[builder(on(...))]` works quite well for attributes that are boolean flags.

However, using it for attributes that can have more states than `true/false` is more complex. For example, `#[builder(default)]` is not just a boolean. It can accept any arbitrary expression: `#[builder(default = 2 + 2)]`. So then there is a question of how `#[builder(on(...))]` should behave when there are multiple `on(...)` directives that match the same member.

There are several ways to approach the problem.

### Prioritization based on pattern specificity

We could obviously say that the pattern `_` is more general than `Box<_>` or `String`. Therefore we could say that if there is `on(String, default = ...)` and `#[builder(_, default)]`, then the `on(String, default = ...)` clause wins and thus its configuration is applied.

[...]

PR #398, opened 2026-08-24
feat: support the default switch in the on(...) clause

### LINKED ISSUE #366: Allow `#[builder(on(Vec, default))]`
I want to add builders to all types generated by [`prost-build`](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.message_attribute).

All fields are optional, but Prost will generate a `Vec` instead of an `Option<Vec>`.
So it would be useful to do `#[builder(on(Vec, default))]`, so those fields are optional too.

ChrisJr404:
Closes #366.

This adds `default` as a switch in the `on(...)` clause, so you can write `#[builder(on(Vec<_>, default))]` to make every matching required member optional and fall back to its `Default` value. It's the boolean form from your comment on the issue, so a custom `default = ...` value is still left to #152.

Members that already have a `#[builder(default = ...)]` or are `Option<T>` are left untouched, so this doesn't change anything for existing builders. I wired it into the same type-pattern matching that `into` and `overwritable` use, added tests to `attr_on.rs`, and documented the attribute on the `on` reference page.

### DIFF
--- a/bon-macros/src/builder/builder_gen/member/named.rs
+++ b/bon-macros/src/builder/builder_gen/member/named.rs
@@ -263,6 +263,8 @@ impl NamedMember {
     }
 
     pub(crate) fn merge_on_config(&mut self, on: &[OnConfig]) -> Result {
+        self.merge_config_default(on)?;
+
         // This is a temporary hack. We only allow `on(_, required)` as the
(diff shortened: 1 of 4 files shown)


Veetaha:
Thanks for the PR, I'll release it soon

